PR #1, opened 2022-04-12
Depth implementation

dc854:


sampsyo:
Great. If that's really how simple the depth algorithm is in its most basic form, then I guess that's good news.

I have two suggestions related to things to keep in the repository, which will be useful for our long-term sanity:

- Let's not include "intermediate" files. In this case, this means `.og` files that we generate from `.gfa` files. The "source of truth" is in the original `.gfa` file, and the `.og` file is derived, so keeping it around is (a) redundant and (b) error-prone if it gets out of date. (This is analogous to the advice never to check in `.o` files that are the result of compiling C programs.) Instead, let's include **some sort of recipe for generating these files**. Maybe a Makefile or even just a README saying how to create them.
- This is a less hard-and-fast rule, but let's *mostly* avoid checking in "other people's data." In this case, that means the `.gfa` files. They are available elsewhere on the Internet, so no need to include them in our repository. Instead, as with the previous bullet point, we can just provide a recipe (either machine-executable or human-readable) for obtaining them.

I'll try to briefly show y'all what I mean in an upcoming PR…

sampsyo:
(BTW, including the extremely tiny `k.gfa` file is probably an exception to the second item, because it is _so_ incredibly tiny…)

PR #3, opened 2022-04-19
Testing Framework

Evan-Williams-02:
Made some modifications to depth.py and added a program to format output of odgi depth


Makefile
```diff
@@ -1,7 +1,8 @@
 GFA_FILES := k.gfa DRB1-3123.gfa
+TEST_FILE := DRB1-3123.gfa
 GFA_URL := https://raw.githubusercontent.com/pangenome/odgi/ebc493f2622f49f1e67c63c1935d68967cd16d85/test
 
-.PHONY: fetch
+.PHONY: fetch test
```
sampsyo, line 5: Just stating the obvious: looks like these Makefile changes aren't used for now.

depth.py
```diff
@@ -13,8 +13,12 @@ def depth(filename):
                 depth_map[name] += 1
             else:
                 depth_map[name] = 1
-    print(depth_map)
+    #print(depth_map)
 
+    with open('python_output.txt', 'w') as f: 
+        depth_items = depth_map.items()
+        for pair in depth_items:
+            f.write(f'{pair[0]} {pair[1]}\n')
```
sampsyo, line 21: I think just `print`ing this this to stdout would be simpler. If you just do `print(f'...')` instead of opening a text file as output, it remains possible to *both* see the output in your console *or* write it to a file. That is, you can do this:

```
$ python3 depth.py something.gfa
```

to see the output with your eyes, or use shell redirection:


```
$ python3 depth.py > output.txt
```

to write it to `output.txt` for testing/comparison purposes. Leaving the output file non-hardcoded will make it easier to manipulate the output when necessary.

sampsyo (commented):
Looks like it's headed in the right direction! I left a few high-level comments within.

I would recommend _not_ committing the output files to the repository. To test, we probably want to compare what the tools output _at that instant_ instead of whenever the outputs were previously generated and checked into the repo. So leaving them out (and possibly even adding them to `.gitignore`) would be a good way to ensure that every test gets "fresh" results to compare.

depth.py
```diff
@@ -8,13 +8,16 @@ def depth(filename):
 
     for path in g.paths:
         for segment in path.segment_names:
-            name = segment.name
+            name = int(segment.name)
```
sampsyo, line 11: Do we know that segment names are always numbers? (If not, then this call to `int(...)` might crash.)

depth.py
```diff
@@ -8,13 +8,16 @@ def depth(filename):
...
             if name in depth_map:
                 depth_map[name] += 1
             else:
                 depth_map[name] = 1
-    print(depth_map)
+    sorted_depth_items = sorted(depth_map.items())
 
+    with open('python_output.txt', 'w') as f: 
+        for pair in sorted_depth_items:
+            f.write(f'{pair[0]} {pair[1]}\n')
```
sampsyo, line 20: I would recommend just printing this to standard output: i.e., just use `print`, not `f.write` (and no need to hard-code an output filename). This will make it easier to use the various scripts for different purposes in the future.

test.sh
```diff
@@ -0,0 +1,42 @@
...
+
+while getopts 'f:ih' OPTION; do
+	case "$OPTION" in
+		f)
+			odgi build -g $OPTARG -o temp.og
+			odgi depth -i temp.og -d > temp_depth.txt
+
+			python3 process.py temp_depth.txt
+
+			python3 depth.py $OPTARG
```
sampsyo, line 20: If you take my suggestion above to avoid hard-coding the output filename, you can do `> python_output.txt` here.

process.py
```diff
@@ -0,0 +1,11 @@
+import sys
```
sampsyo, line 1: This file could use some comments describing what it does.

process.py
```diff
@@ -0,0 +1,11 @@
...
+
+with open(sys.argv[1], 'r') as f: 
+    contents = f.readlines()
+    data = []
+    for line in contents:
+        data.append(line.split())
+
+with open('odgi_output.txt', 'w') as f2:
+    for i in range(1, len(data)):
+        f2.write(f'{data[i][0].strip()} {data[i][1].strip()}\n')
```
sampsyo, line 11: I'd also avoid hard-coding the output filename here too. Just using a normal `print` will make this print to the standard output stream, making things more flexible.

process.py
```diff
@@ -0,0 +1,11 @@
+import sys
+
+with open(sys.argv[1], 'r') as f:
```
sampsyo, line 3: This isn't critical, but for bonus points, you can avoid needing to take a file name as input by just reading from standard input… that is, read directly from `sys.stdin`. Then you would invoke this script like:

```
python3 process.py < input.txt > output.txt
```

to read data from `input.txt` and write it to `output.txt`.

sampsyo (commented):
Looking good! Added a few comments—one high-level suggestion is to make the various command-line tools read from stdin and write to stdout, which will avoid hard-coding specific filenames to read/write. (This will make the tools easier for humans to use and also make them more flexible as we build more tooling around them.)

PR #4, opened 2022-04-24
Depth with file

dc854:
- Removed dependency on `gfapy`
- `format_gfa.py` will now turn a `.gfa` file to a flat file format or return a list of segments in all path
- `depth.py` changed to use the list of segments from `format_gfa.py`
- `depth2.py` will generate a `.json` file if the provided a `.gfa` file when called


sampsyo (commented):
Looking promising! Here are a few suggestions. I have some ideas for refactoring to make the code more reusable, but I would merge this immediately and do those in a separate PR.

For this PR, please just do these minor things:

* Please add comments and docstrings so we can understand the code a bit better.
* Please remove the `__pycache__` directory (and probably add it to our `.gitignore` to avoid accidentally adding it in the future).

PR #10, opened 2022-09-29
Update README.md

susan-garry:
- Fix bioconda link
- Include a section with all of the code to run a hardware accelerator
- Provide more details on setting up `calyx` and `odgi`
- Include an example for parsing data where a file is passed to the -a flag
- Divide setting up dependencies and generating an accelerator into two difference sections

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,21 +14,21 @@ Then follow the instructions below to set up `calyx` and `odgi`.
 ### Installing Dependencies
 
 #### Calyx
-Follow these [instructions](https://docs.calyxir.org/) to install calyx. You will need to configure the calyx driver, `fud`. We recommend using the native calyx interpreter, so once `fud` is set up, run
+Follow these [instructions](https://docs.calyxir.org/) to install calyx. You will need to complete the Installing the Commandline Driver section, but can skip Running the Core Tests. We recommend using the native calyx interpreter, so once `fud` is set up, run
 ```
-fud config stages.interpreter $ABSOLUTE_PATH_TO_CALYX_REPO/target/debug/interp
+fud config stages.interpreter <full path to Calyx repository>/target/debug/interp
 ```
-where `$ABSOLUTE_PATH_TO_CALYX_REPO` is the absolute path to the root directory. For example, if you downloaded calyx in `/Users/username/project`, you would run `fud config stages.interpreter /Users/username/project/target/debug/interp`.
+where `<full path to Calyx repo` is the absolute path to the root directory. For example, if you downloaded calyx in `/Users/username/project`, you would run `fud config stages.interpreter /Users/username/project/calyx/target/debug/interp`.
 
 #### Odgi
 
-You will need to install the python bindings for [odgi]. Instructions for installing odgi can be found [here](https://odgi.readthedocs.io/en/latest/rst/installation.html). If you compile odgi from its source, you will need to [edit your python path](https://odgi.readthedocs.io/en/latest/rst/binding/usage.html) to use the python bindings.
+You will need to install the python bindings for [odgi]. Instructions for installing odgi can be found [here](https://odgi.readthedocs.io/en/latest/rst/installation.html). Installing odgi via `bioconda` seems to be the most straightforward option. If you instead compile odgi from its source, you will need to [edit your python path](https://odgi.readthedocs.io/en/latest/rst/binding/usage.html) to use the python bindings. 
 
-To verify that you have the python bindings, open up a python shell and try `import odgi`. If this doesn't work, you can try a different method of installation, or you can download the `.so` files from [bioconda](https://anaconda.org/bioconda/odgi/files) for the version of python you are running (`python3 --version`) and add them to your `PYTHONPATH` instead.
+To verify that the python bindings are working, open up a python shell and try `import odgi`. If this doesn't work, you can also download the `.so` files from [bioconda](https://anaconda.org/bioconda/odgi/files) for the version of python you are running and add them to your `PYTHONPATH`. For example, if `python --version` is 3.7, fetch `odgi...py37....tar.bz2`. You will have to extract the `.so` files by unzipping the `.tar.bz2` file.
 
 ### Generating an Accelerator
 
-To generator and run an accelerator to compute the node depth for `k.og`, first navigate to the root directory of this repository. Then run
+Take node depth as an example. To generate and run a node depth accelerator for `k.og`, first navigate to the root directory of this repository. Then run
 ```
 make fetch
 make test/k.og
--- a/README.md
+++ b/README.md
@@ -14,17 +14,19 @@ Then follow the instructions below to set up `calyx` and `odgi`.
 ### Installing Dependencies
 
 #### Calyx
-Follow these [instructions](https://docs.calyxir.org/) to install calyx. You will need to complete the Installing the Commandline Driver section, but can skip Running the Core Tests. We recommend using the native calyx interpreter, so once `fud` is set up, run
+Follow these [instructions](https://docs.calyxir.org/) to install calyx. You must complete the [first](https://docs.calyxir.org/#compiler-installation) and [third] (https://docs.calyxir.org/#installing-the-command-line-driver) sections, but feel free to skip the second. We recommend using the native calyx interpreter, so once `fud` is set up, run
 ```
 fud config stages.interpreter <full path to Calyx repository>/target/debug/interp
 ```
-where `<full path to Calyx repo` is the absolute path to the root directory. For example, if you downloaded calyx in `/Users/username/project`, you would run `fud config stages.interpreter /Users/username/project/calyx/target/debug/interp`.
+where `<full path to Calyx repository` is the absolute path to the root directory. For example, if you downloaded calyx in `/Users/username/project`, you would run `fud config stages.interpreter /Users/username/project/calyx/target/debug/interp`.
 
 #### Odgi
 
-You will need to install the python bindings for [odgi]. Instructions for installing odgi can be found [here](https://odgi.readthedocs.io/en/latest/rst/installation.html). Installing odgi via `bioconda` seems to be the most straightforward option. If you instead compile odgi from its source, you will need to [edit your python path](https://odgi.readthedocs.io/en/latest/rst/binding/usage.html) to use the python bindings. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,38 +7,70 @@ It is very much a work in progress.
...
+#### Odgi
+
+You will need to install the python bindings for [odgi]. Instructions for installing odgi can be found [here](https://odgi.readthedocs.io/en/latest/rst/installation.html). If you compile odgi from its source, you will need to [edit your python path](https://odgi.readthedocs.io/en/latest/rst/binding/usage.html) to use the python bindings.
+
+To verify that you have the python bindings, open up a python shell and try `import odgi`. If this doesn't work, you can try a different method of installation, or you can download the `.so` files from [bioconda](https://anaconda.org/bioconda/odgi/files) for the version of python you are running (`python3 --version`) and add them to your `PYTHONPATH` instead.
 
-Then, to generate a node depth hardware accelerator in calyx, type one of the following commands:
+### Generating an Accelerator
+
+To generator and run an accelerator to compute the node depth for `k.og`, first navigate to the root directory of this repository. Then run
```
anshumanmohan, line 31: minor: "generate and run"

README.md
```diff
@@ -7,38 +7,70 @@ It is very much a work in progress.
...
+```git clone https://github.com/cucapra/calyx-pangenome.git```
+Then follow the instructions below to set up `calyx` and `odgi`.
 
-You will need to fetch some useful inputs by typing `make fetch`.
-This downloads some simple [GFA][] data files into the `./test` directory.
+### Installing Dependencies
 
-If you want to build odgi graph files from the GFA files, try something like `make k.og`.
+#### Calyx
+Follow these [instructions](https://docs.calyxir.org/) to install calyx. You will need to configure the calyx driver, `fud`. We recommend using the native calyx interpreter, so once `fud` is set up, run
```
anshumanmohan, line 17: I think it may pay to say that they can skip the Running Core Tests portion, and should get until `fud check`.

README.md
```diff
@@ -7,38 +7,70 @@ It is very much a work in progress.
...
+```
+fud config stages.interpreter $ABSOLUTE_PATH_TO_CALYX_REPO/target/debug/interp
```
anshumanmohan, line 19: To be honest I find Calyx's `config` steps very brittle. I was trying variations of this command to see if I could simplify it, and I now think my setup is messed up. Deleting and re-cloning Calyx has not worked. Regardless, I think it would pay to follow Calyx's style of notation, <full path to Calyx repository>. Now, in the next line, you say that I should actually stop one level _outside_ the Calyx repository. Just wanted to check you on that...

README.md
```diff
@@ -7,38 +7,70 @@ It is very much a work in progress.
...
+```
+where `$ABSOLUTE_PATH_TO_CALYX_REPO` is the absolute path to the root directory. For example, if you downloaded calyx in `/Users/username/project`, you would run `fud config stages.interpreter /Users/username/project/target/debug/interp`.
+
+#### Odgi
+
+You will need to install the python bindings for [odgi]. Instructions for installing odgi can be found [here](https://odgi.readthedocs.io/en/latest/rst/installation.html). If you compile odgi from its source, you will need to [edit your python path](https://odgi.readthedocs.io/en/latest/rst/binding/usage.html) to use the python bindings.
+
+To verify that you have the python bindings, open up a python shell and try `import odgi`. If this doesn't work, you can try a different method of installation, or you can download the `.so` files from [bioconda](https://anaconda.org/bioconda/odgi/files) for the version of python you are running (`python3 --version`) and add them to your `PYTHONPATH` instead.
```
anshumanmohan, line 27: Feels like we're leaving them in the lurch, asking them to flail around and maybe also fix versions. Maybe we'll learn more ourselves in the weeks to come, but eventually it would be nice to say, "Here is the odgi install guide; they give three options but we we recommend that you chose [the Bioconda route](https://odgi.readthedocs.io/en/latest/rst/installation.html#bioconda)." Or something like that.

For the .so file, it may be worth giving a tiny example. So like, "If the result of `python3 --version` is 3.7, fetch `odgi...py37....tar.bz2`.

README.md
anshumanmohan: I don't think it's clear that node depth is merely a working example. May be good to just say that briefly.

anshumanmohan (changes requested):
Thanks, Susan! Looking so good. I added some minor comments inline (please let me know if you can't see them...) and I'd love to chat sometime to polish this off and merge it in. :)

README.md
susan-garry: Good catch! The next line should be `Users/username/project/calyx/target/debug/interp`. I will also change the notation style.

I also had some issues with configuring calyx, and the issue ended up being with the config file. On havarti, the config file is located at `~/.config/fud/config.toml` for me. You could try deleting this file, which should effectively revert you back to the defaults, and see if it fixes the issues you've been running into?

If you're not using havarti or if this doesn't fix the problem, shoot me a message!

README.md
```diff
@@ -14,21 +14,21 @@ Then follow the instructions below to set up `calyx` and `odgi`.
...
 
 #### Calyx
-Follow these [instructions](https://docs.calyxir.org/) to install calyx. You will need to configure the calyx driver, `fud`. We recommend using the native calyx interpreter, so once `fud` is set up, run
+Follow these [instructions](https://docs.calyxir.org/) to install calyx. You will need to complete the Installing the Commandline Driver section, but can skip Running the Core Tests. We recommend using the native calyx interpreter, so once `fud` is set up, run
 ```
-fud config stages.interpreter $ABSOLUTE_PATH_TO_CALYX_REPO/target/debug/interp
+fud config stages.interpreter <full path to Calyx repository>/target/debug/interp
 ```
-where `$ABSOLUTE_PATH_TO_CALYX_REPO` is the absolute path to the root directory. For example, if you downloaded calyx in `/Users/username/project`, you would run `fud config stages.interpreter /Users/username/project/target/debug/interp`.
+where `<full path to Calyx repo` is the absolute path to the root directory. For example, if you downloaded calyx in `/Users/username/project`, you would run `fud config stages.interpreter /Users/username/project/calyx/target/debug/interp`.
```
anshumanmohan, line 21: where `<full path to Calyx repository>` is the absolute path to the root directory of the calyx project.

README.md
```diff
@@ -14,21 +14,21 @@ Then follow the instructions below to set up `calyx` and `odgi`.
 ### Installing Dependencies
 
 #### Calyx
-Follow these [instructions](https://docs.calyxir.org/) to install calyx. You will need to configure the calyx driver, `fud`. We recommend using the native calyx interpreter, so once `fud` is set up, run
+Follow these [instructions](https://docs.calyxir.org/) to install calyx. You will need to complete the Installing the Commandline Driver section, but can skip Running the Core Tests. We recommend using the native calyx interpreter, so once `fud` is set up, run
```
anshumanmohan, line 17: How about something like this?
> You must complete the [first](https://docs.calyxir.org/#compiler-installation) and [third](https://docs.calyxir.org/#installing-the-command-line-driver) sections, but feel free to skip the second.

README.md
susan-garry: Yep. Good catch!

README.md
```diff
@@ -14,21 +14,21 @@ Then follow the instructions below to set up `calyx` and `odgi`.
...
 
-To verify that you have the python bindings, open up a python shell and try `import odgi`. If this doesn't work, you can try a different method of installation, or you can download the `.so` files from [bioconda](https://anaconda.org/bioconda/odgi/files) for the version of python you are running (`python3 --version`) and add them to your `PYTHONPATH` instead.
+To verify that the python bindings are working, open up a python shell and try `import odgi`. If this doesn't work, you can also download the `.so` files from [bioconda](https://anaconda.org/bioconda/odgi/files) for the version of python you are running and add them to your `PYTHONPATH`. For example, if `python --version` is 3.7, fetch `odgi...py37....tar.bz2`. You will have to extract the `.so` files by unzipping the `.tar.bz2` file.
```
anshumanmohan, line 27: I think this is getting hair enough that we need a little list, or even a section, to cover it.

anshumanmohan: 1. Check Python3 version. Let's say it is 3.9.12.
2. Make `odgi` directory.
3. Download the appropriate tarball (let's say "download" or "`curl`" and not "fetch"? :) )
4. Untar it, and explore under `/lib/python3.9/site-packages/` to ensure that `odgi.cpython*.so` is there. If it is elsewhere, make note of the location and substitute in the next step.
5. Add this to your PYTHONPATH: `export PYTHONPATH=...odgi/lib/python3.9/site-packages/`
6. Explore under `/usr/lib/x86_64-linux-gnu/` to ensure that `libjemalloc.so.2` is there. If it is not, search under  `/lib/x86_64-linux-gnu/` and substitute in the next step.
7. Run `export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libjemalloc.so.2`
8. Run `python` and then `import odgi`.

susan-garry: Do you remember which command you used to download the tarball? I tried curl just now but couldn't get it to work for me, so I ended up downloading the tarball to my computer and using `scp`.

anshumanmohan:
Looking great, Susan. Thanks for taking the tip re: corresponding lists of code; I think that worked out well. One last nitpick and we'll hit merge:
`<full path to Calyx repository` is missing a closing `>`

README.md
anshumanmohan: Ah, this may just be a matter of using the right incantations!
1. `wget internet.address/file.zip` will get `file.zip`, with that name, on your machine.
2. `curl internet.address/file.zip` will give you a silly wall of text, but `curl internet.address/file.zip --output localfile.zip` will get `file.zip`, now renamed `localfile.zip`, on your machine.

FWIW, I was just able to download Odgi for Python 3.9 onto Havarti using both ways. 

Does this help, or is it the case that you already knew this and were facing a deeper issue?

README.md
anshumanmohan: There's a minor bug (caused by an errant whitespace) in the current version. The [third](example.com) link is broken.

README.md
susan-garry: Ah, you are correct, I was using the wrong incantation. Thank you for clearing that up!

PR #11, opened 2022-10-12
Work on Docker image

anshumanmohan:
This PR introduces Docker support. The main deliverable is the Dockerfile we will use to build the image, which we will then push to GHCR. Let's merge this branch in once we are happy with the Docker image that the Dockerfile creates.

I have already built and pushed an image based on the current Dockerfile; this is temporarily hosted on Dockerhub. Pull and run it with
```
docker pull anshumanmohan/pollen
docker run -it anshumanmohan/pollen /bin/bash
```

To make changes to the image, navigate so that this Dockerfile in your working directory. Then
```
docker build -t username/pollen .
docker run -it username/pollen /bin/bash
```
Note the period at the end of the first command.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,10 +6,6 @@
 
 # Start with latest Calyx image
 FROM ghcr.io/cucapra/calyx:latest
-# TODO-AM: 
-#    part of the point is to be "stable".
-#    pick a version of this and everything else,
-#    updating only when stable?
 
 # return to root directory
 WORKDIR /root
@@ -39,9 +35,6 @@ RUN git clone https://github.com/cucapra/calyx-pangenome.git
 # build:
 WORKDIR /root/calyx-pangenome
 RUN make fetch
-# TODO: 
-# RUN make test-all
-# many failures; not sure why 
 
 # return to root directory
 WORKDIR /root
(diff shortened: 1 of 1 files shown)


anshumanmohan:
I imagine the improvements we'll want will include
1. Getting all the flavors of `make` to work correctly.
2. Doing some hacking to get resource estimation (via `fud`'s remote facility) working.

A (short) list of things that *do* work at present, and we should try not to break as we make progress:
1. `fud` works from the command line.
2. `odgi` works from the command line.
3. `import odgi` works from inside a Python environment.

susan-garry:
This is a great idea! The current problem that I am running into is that the `calyx.py_ast` module is not available from our docker image, even though it is available in the base image (the calyx docker image). At the moment, I don't know what is causing this

susan-garry:
Okay, I fixed the import issue (`odgi` was only visible from the root directory; `calyx.py_ast` seems not to have had any issues), and all of the tests now pass when we run `make test` inside the docker.

The only thing left to do, I believe, is figure out how to get resource estimation working (maybe try `ssh`ing into the server that the docker is running on?)

anshumanmohan:
I've updated the imagine on Dockerhub so it reflects the changes above.

anshumanmohan:
Quick update, I've gotten through the GHCR crud and now we have a nice Packages section on the repo's landing page.

PR #14, opened 2022-10-16
Update readme

anshumanmohan:
At the risk of gilding the lily, I've made another pass at the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,15 +29,14 @@ Follow these [instructions](https://docs.calyxir.org/) to install calyx. You mus
 ```
 fud config stages.interpreter <full path to calyx repository>/target/debug/interp
 ```
-where `<full path to calyx repository>` is the absolute path to the calyx root directory, ending in `/calyx/`. 
 
 #### Odgi
 
-You will need to install the python bindings for [odgi]. Instructions for installing odgi can be found [here](https://odgi.readthedocs.io/en/latest/rst/installation.html). Installing odgi via `bioconda` seems to be the most straightforward option. If you instead compile odgi from its source, you will need to [edit your python path](https://odgi.readthedocs.io/en/latest/rst/binding/usage.html) in order to use the python bindings.
+Installing odgi [via bioconda](https://odgi.readthedocs.io/en/latest/rst/installation.html#bioconda) seems to be the most straightforward option. If you instead [compile odgi from source](https://odgi.readthedocs.io/en/latest/rst/installation.html#building-from-source), you will need to [edit your python path](https://odgi.readthedocs.io/en/latest/rst/binding/usage.html) in order to use the python bindings.
 
 To verify that odgi is installed and the python bindings are working, open up a python shell and try `import odgi`. If this works, move on to the next section.
 
-We have encountered two gotchas when installing odgi: a version clash with python, and an issue with odgi's memory manager. Below we describe what we think is a complete installation of odgi that circumvents both of these issues.
+We have encountered two gotchas when installing odgi: a version clash with python, and an issue with odgi's memory manager. Below we describe what we think is a complete installation of odgi that addresses both of these issues.
 
 1. Check your python version with `python --version`. We use python 3.9.12 for the rest of this example.
 2. Run `mkdir odgi-py; cd odgi-py`.
@@ -46,7 +45,7 @@ We have encountered two gotchas when installing odgi: a version clash with pytho
 5. Add this to your `PYTHONPATH` with `export PYTHONPATH=<full path to odgi-py>/lib/python3.9/site-packages/`.
 6. Preload `jemalloc`: explore under `/usr/lib/x86_64-linux-gnu/` to ensure that `libjemalloc.so.2` is there. If it is not, search under `/lib/x86_64-linux-gnu/` and substitute in the next step.
 7. Run `export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libjemalloc.so.2`.
-8. Run `python` and then `import odgi`.
+8. Open up a python shell and try `import odgi`.
 
 ### Generating an Accelerator
 
@@ -60,7 +59,7 @@ fud exec depth.futil --to interpreter-out -s verilog.data depth.data > depth.txt
 python3 parse_data.py -di temp.txt
 ```
 
-What just happened? Below we walk through the six commands we issued above, pointing out the other options that we could have used.
+What just happened? Below, we walk through the six commands we issued above, pointing out the other options that we could have used.
 
 First, `make fetch` downloads some [GFA][] data files into the `./test` directory.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,43 +54,41 @@ fud exec depth.futil --to interpreter-out -s verilog.data depth.data > depth.txt
...
 
 The commands use the hardware parameters as follows:
-1. Uses default hardware parameters
-2. Automatically infers the hardware parameters from a `.og` file
-3. Takes the hardware parameters as input.
+1. Uses default hardware parameters.
+2. Takes the hardware parameters as input.
+3. Automatically infers the hardware parameters from a `.og` file.
 
 Automatically inferred parameters take precedence over manually specified ones, and a subset of parameters may be specified. For example, `python3 calyx_depth.py -a test/k.og -n=1` will infer `MAX_STEPS` and `MAX_PATHS` from `test/k.og`, but the resulting accelerator can only handle one node.
```
anshumanmohan, line 80: @susan-garry sorry if I'm being silly, but this sentence seems a little contradictory to me. In English you say that the automatic inference takes precedence, but then the example seems to demonstrate the opposite.

README.md
```diff
@@ -53,43 +54,41 @@ fud exec depth.futil --to interpreter-out -s verilog.data depth.data > depth.txt
...
+Second, `make test/*.og` builds the odgi graph files from those GFA files.
 
-Then, `python3 calyx_depth.py -o depth.futil` generates the hardware accelerator and writes it to a file named `depth.futil`. The commands to generate a node depth hardware accelerator in calyx include:
+Third, we generate the hardware accelerator and write it to a file named `depth.futil`. The commands to generate a node depth hardware accelerator in calyx include:
 
 1. `python3 calyx_depth.py -o depth.futil`
-2. `python3 calyx_depth.py -a <filename> -o depth.futil`
-3. `python3 calxy_depth.py -n=MAX_NODES -e=MAX_STEPS -p=MAX_PATHS -o depth.futil`
+2. `python3 calxy_depth.py -n=MAX_NODES -e=MAX_STEPS -p=MAX_PATHS -o depth.futil`
+3. `python3 calyx_depth.py -a <filename.og> -o depth.futil`
```
anshumanmohan, line 73: Just an experiment, adding the expected extensions of the various files. If you like it, we can do it all over. If you hate it, I believe this is the only spot I added it!

README.md
```diff
@@ -7,43 +7,49 @@ Pangenome Graph Queries in Calyx
...
+git clone https://github.com/cucapra/pollen.git
+```
+and then follow the instructions below to set up our dependencies, `calyx` and `odgi`.
 
-### Installing Dependencies
 
 #### Calyx
-Follow these [instructions](https://docs.calyxir.org/) to install calyx. You must complete the [first](https://docs.calyxir.org/#compiler-installation) and [third](https://docs.calyxir.org/#installing-the-command-line-driver) sections, but feel free to skip the second. We recommend using the native calyx interpreter, so once `fud` is set up, run
+
+Follow these [instructions](https://docs.calyxir.org/) to install calyx. You must complete the [first](https://docs.calyxir.org/#compiler-installation) and [third](https://docs.calyxir.org/#installing-the-command-line-driver) sections, but feel free to skip the second. We recommend using the native calyx interpreter, so once `fud check` has passed, run
```
susan-garry, line 28: I think saying "once `fud check`" has passed is a bit misleading, because it is expected that many stages won't be installed and when you run `fud check`, roughly the same number of checks will 'fail' as will 'pass.' I think as long as `fud check` runs, things should be good. We could say something like
> once `fud check` has passed (it is expected that some tools will not be available), run

README.md
susan-garry: That's probably a good idea to add the expected file extensions. I can go through and change it if you don't get around to it first

susan-garry: Just curious, what was the reason for swapping the order of 2 and 3?

README.md
```diff
@@ -53,43 +59,41 @@ fud exec depth.futil --to interpreter-out -s verilog.data depth.data > depth.txt
...
+3. Automatically infers the hardware parameters from a `.og` file.
 
-Automatically inferred parameters take precedence over manually specified ones, and a subset of parameters may be specified. For example, `python3 calyx_depth.py -a test/k.og -n=1` will infer `MAX_STEPS` and `MAX_PATHS` from `test/k.og`, but the resulting accelerator can only handle one node.
+Automatically-inferred parameters take precedence over manually specified ones, and a subset of parameters may be specified. For example, `python3 calyx_depth.py -a test/k.og -n=1` will infer `MAX_STEPS` and `MAX_PATHS` from `test/k.og`, but the resulting accelerator can only handle one node.
 
-To run the hardware accelerator, we need to generate some input using one of the following commands:
+Fourth, we need to generate some input from our odgi file. This is what we will feed to the hardware accelerator. The following variations all accomplish this:
 
 1. `python3 parse_data.py <filename> -o depth.data`
 2. `python3 parse_data.py <filename> -a <filename2> -o depth.data`
```
susan-garry, line 84: 2 and 3 were swapped above. For consistency, we should have them in the same order in both places

README.md
```diff
@@ -53,43 +59,41 @@ fud exec depth.futil --to interpreter-out -s verilog.data depth.data > depth.txt
...
 The commands use the hardware parameters as follows:
-1. Uses default hardware parameters
-2. Automatically infers the hardware parameters from a `.og` file
-3. Takes the hardware parameters as input.
+1. Uses default hardware parameters.
+2. Takes the hardware parameters as input.
+3. Automatically infers the hardware parameters from a `.og` file.
 
-Automatically inferred parameters take precedence over manually specified ones, and a subset of parameters may be specified. For example, `python3 calyx_depth.py -a test/k.og -n=1` will infer `MAX_STEPS` and `MAX_PATHS` from `test/k.og`, but the resulting accelerator can only handle one node.
+Automatically-inferred parameters take precedence over manually specified ones, and a subset of parameters may be specified. For example, `python3 calyx_depth.py -a test/k.og -n=1` will infer `MAX_STEPS` and `MAX_PATHS` from `test/k.og`, but the resulting accelerator can only handle one node.
```
susan-garry, line 79: Nitpick, but is it common to hyphenate automatically-inferred? I don't know why "automatically inferred" would be hyphenated, but "manually specified" wouldn't be.

README.md
susan-garry: This is correct. I definitely edited this, but maybe it was on a different branch because it clearly didn't make the final cut. "Manually specified" and "automatically inferred" should be swapped now

README.md
susan-garry: That's probably a good idea to add the expected file extensions. I can go through and change it if you don't get around to it first

anshumanmohan: Heh, no hard reason! I guess in my head it flows better this way, since I view the supplying of `filename.og` to be a more heavyweight way to do it. Happy to swap it back!

README.md
anshumanmohan: Thank you, done!

README.md
anshumanmohan: Thanks for the catch; done!

README.md
anshumanmohan: To be honest I'm not sure myself, but I massaged the text to get away from the issue :)

susan-garry: For future reference, I was going through the capra style guidelines to see if there was anything about writing github PRs, and there's a section about [when to use hyphens](https://capra.cs.cornell.edu/styleguide/#hyphen). (It says don't do it when the word ends in _-ly_).

README.md
susan-garry: That is fair; it definitely takes more time for the code to infer the parameters. In my head I was viewing it as more of a burden on the user to have to specify each parameter

README.md
susan-garry: Looks great!

README.md
susan-garry: File extensions updated!

README.md
susan-garry: Although in this case, I think I prefer your latest version!

README.md
anshumanmohan: Heh so maybe I would not have screwed them up after all...

PR #22, opened 2023-02-16
A minor expansion of mygfa to compute node_steps

anshumanmohan:
mygfa.py now computes the NodeSteps<sub>s</sub> table described in https://github.com/cucapra/notes/blob/main/Pollen/relational.md#indexes. As Adrian suggests in the note, `odgi depth` is now a cinch. 

Just as a sanity check, here's a sketch of `k.gfa`.
IMG_0041

This produces the following node depth table, which I think is correct.
Screen Shot 2023-02-15 at 7 12 04 PM


mygfa.py
```diff
@@ -25,6 +25,21 @@ def parse(cls, fields: List[str]) -> "Segment":
         _, name, seq = fields[:3]
         return Segment(name, seq)
 
+    """Compact any "runs" of N down to a single N."""
+    def crush_n(self):
```
sampsyo, line 29:
```suggestion
    def crush_n(self) -> "Segment":
```

To make it clear what this returns, (i.e., it doesn't mutate `self`).

mygfa.py
```diff
@@ -120,6 +135,7 @@ def parse(cls, fields: List[str]) -> "Path":
     def __str__(self):
         return '\t'.join([
             "P",
+            self.name,
```
sampsyo, line 138: Oops; thanks.

mygfa.py
```diff
@@ -161,6 +177,11 @@ def parse(cls, infile: TextIO) -> "Graph":
 
         return graph
 
+    def crush_n(self):
```
sampsyo, line 180:
```suggestion
    def crush_n(self) -> "Graph":
```

Similarly.

mygfa.py
```diff
@@ -161,6 +177,11 @@ def parse(cls, infile: TextIO) -> "Graph":
...
+        crushed_segments = \
+            {name: Segment.crush_n(seg) for name, seg in self.segments.items()}
```
sampsyo, line 182:
```suggestion
            {name: seg.crush_n() for name, seg in self.segments.items()}
```

mygfa.py
```diff
@@ -170,6 +191,36 @@ def emit(self, outfile: TextIO):
...
+    for path in graph.paths.values():
+        for id, (seg_name, seg_orient) in enumerate(path.segments):
+            crossings[seg_name].append((path.name, id, seg_orient))
+
+    return crossings
+    # Feel free to
+    # print(crossings)
+    # to test it by eye, but it's pretty ugly.
+    # This is essentially the (p,i,d) table imagined here:
+    # https://github.com/cucapra/notes/blob/main/Pollen/relational.md#indexes
```
sampsyo, line 212: Probably no need for the hint about printing IMO.

mygfa.py
```diff
@@ -170,6 +191,36 @@ def emit(self, outfile: TextIO):
...
+    # again as observed in the note.
+    for (segment, crossings) in node_steps(graph).items():
+        print('\t'.join([segment, str(len(crossings))]))
+
+
 if __name__ == "__main__":
     graph = Graph.parse(sys.stdin)
+    graph = Graph.crush_n(graph)
     graph.emit(sys.stdout)
+    # node_depth(graph)
```
sampsyo, line 226: If it's OK, I would leave these out of the `__main__` thing. In a future phase, maybe we can move the commands to a different module where they have a proper CLI.

mygfa.py
```diff
@@ -170,6 +191,36 @@ def emit(self, outfile: TextIO):
...
+    # print(crossings)
+    # to test it by eye, but it's pretty ugly.
+    # This is essentially the (p,i,d) table imagined here:
+    # https://github.com/cucapra/notes/blob/main/Pollen/relational.md#indexes
+
+
+def node_depth(graph):
+    # Here I show that node_depth is just the cardinality of the above,
+    # again as observed in the note.
+    for (segment, crossings) in node_steps(graph).items():
```
sampsyo, line 218:
```suggestion
    for segment, crossings in node_steps(graph).items():
```

sampsyo (changes requested):
Cool! Some Python-y suggestions within.

As one comment implies, it would be cool (not in this PR) to move these various processing commands to a different Python module from the data structures.

anshumanmohan:
This branch has a few more commits than the PR proposed, but I'm merging it and moving on. The additional commits are towards
- rudimentary testing (with diff instead of turnt)
- refactoring code into modules
- implementing `odgi degree`

PR #29, opened 2023-03-18
Slow odgi edgy

anshumanmohan:
Currently a draft, but this PR will add a few edge-aware odgi functions to slow-odgi


Makefile
```diff
@@ -50,9 +50,10 @@ test-slow-matrix: og
 	-turnt --save --env matrix_oracle test/*.og
 	turnt --diff -v --env matrix_test test/*.gfa
 
-test-slow-validate: og
-	-turnt --save --env validate_oracle test/*.gfa
-	turnt --diff -v --env validate_test test/*.gfa
+test-slow-validate: fetch
+	-turnt --save --env validate_setup/*.gfa
```
anshumanmohan, line 54: I need to learn more about turnt, but I can't get this little chunk to run. What I want is:
- Run `perturb` to generate, for each file.gfa, a new temp.file.gfa. This temp.file.gfa will be missing a few links, thus giving validate something to work with.
- Run `odgi validate` on temp.file.gfa, saving the result of this as file.out
- Run slow-odgi's validate on temp.file.gfa, comparing against file.out

However, it seems to me that the test-slow-validate stanza is not running at all right now.

test/turnt.toml
```diff
@@ -54,10 +54,14 @@ command = "odgi matrix -i {filename}"
...
 command = "python3 ../slow_odgi/matrix.py < {filename}"
 
+[envs.validate_setup]
+binary = true
+command = "python3 ../slow_odgi/perturb.py < {filename} > temp.{filename}"
+
 [envs.validate_oracle]
 binary = true
-command = "odgi validate -i {filename}"
+command = "odgi validate -i temp.{filename} 2>temp.txt; cat temp.txt"
```
anshumanmohan, line 63: `odgi validate` puts its output into stderr. What I really want is to direct stderr to stdout, which will itself go to filename.out thanks to turnt. However, this isn't working. I think it has to do with the order, as suggested by [these folks](https://linuxize.com/post/bash-redirect-stderr-stdout/#redirecting-stderr-to-stdout).

My workaround: I'm throwing stderr into a temp file, and then using `cat` to essentially echo it back to turnt, which saves it to filename.out.

test/perturb.sh
anshumanmohan, line 1: Ugly, but I've used a quick sh script to do what I was hoping turnt could do

Makefile
anshumanmohan: I've done up a quick band-aid solution here: [Temp fix to generate interesting cases for validate](https://github.com/cucapra/pollen/pull/29/commits/6a909ba88090ca9eddaa7c3f8ed1642513090eb3) but I'd like to actually get it right using turnt

anshumanmohan:
This PR adds functionality for `odgi matrix` and `odgi validate`. These methods, along with `odgi degree`, are the primary users of edges (aka links). Other than that, edges can mostly be given second-class treatment: paths carry the same information in a nicer format, assuming we do not worry about unused edges that were not on any path to begin with.

These three algorithms all benefit from the same precomputation. In one pass through the graph's links, I construct the two dicts `ins` and `outs`. They have the same type; one holds in-edges and one holds out-edges.

```
Dict[Tuple[str, bool], List[Tuple[str, bool]]]
# where 
# key: (segment name, orientation)               (my details)
# value: list of (segment name, orientation)     (my neighbors' details)
```

```
for link in graph.links:
    ins[(link.to, link.to_orient)].append((link.from_, link.from_orient))
    outs[(link.from_, link.from_orient)].append((link.to, link.to_orient))
```
The three algorithms fall out straightforwardly from these precomputed dicts.

anshumanmohan:
Unfortunately, `matrix` and `validate` both have one case each where they do not diff out correctly against the odgi oracle.
- `matrix` has a simple normalization issue. In the sparse matrix representation of LPA.gfa, the order of nodes I present differs from the order that odgi presents in two instances (two rows of the matrix). The information itself is correct.
- In `validate`, my version reports a few more missing links than odgi does. This is only for the graph DRB1-3123.gfa. 

I will investigate both of these further!

Re: testing and SE, there is another silly issue: I am occasionally forced to create temp files, and those have to have the .gfa extension in order for odgi to accept them as inputs. Then, because they have the .gfa extension, turnt tries to run tests on these temp files as though they were bona fide test files. Silly and a bit gross, and I'd appreciate any thoughts about cleaning up this workflow.

anshumanmohan:
Fixed the normalization issue in `matrix` easily enough. 

The issue with `validate` was more subtle. When the GFA has the link  `L n1 + n2 +` and a path needs to traverse `...n2-, n1-,...`, odgi wants us to say that _this is okay_. I had been saying that this is not okay (and requiring a link that actually looked like `L n2 - n1 -`). I have weakened my check and I diff out correctly versus `odgi validate` now,  but really I think this is kinda bananas and goes to show how silly edges are.

sampsyo:
Wow; hella awesome! I'm glad this edge stuff didn't present _overly_ much trouble!

> These three algorithms all benefit from the same precomputation. In one pass through the graph's links, I construct the two dicts `ins` and `outs`. They have the same type; one holds in-edges and one holds out-edges.

FWIW, I might summarize this by calling it the "adjacency list" representation of a graph.

> I am occasionally forced to create temp files, and those have to have the .gfa extension in order for odgi to accept them as inputs. Then, because they have the .gfa extension, turnt tries to run tests on these temp files as though they were bona fide test files. Silly and a bit gross, and I'd appreciate any thoughts about cleaning up this workflow.

Ah, that is annoying: to summarize, both odgi and Turnt are extension-sensitive, so they are competing for the namespace here. One option might be to build up a teensy bit of infrastructure to put those temporary files into a temporary directory… although using [mktemp(1)](https://linux.die.net/man/1/mktemp) to make a file with a special extension is going to be pretty annoying. I wonder if we might be happiest just implementing something like https://github.com/cucapra/turnt/issues/23 in Turnt?

slow_odgi/perturb.py
```diff
@@ -0,0 +1,18 @@
+import sys
+import mygfa
+import random
+
+def pop90percent(l):
+    origlen = len(l)
+    while len(l) > .1 * origlen:
+        l.pop(random.randrange(len(l)))
```
sampsyo, line 8: FWIW, the [random](https://docs.python.org/3/library/random.html) module has something pretty similar to this:

```py
l[:] = random.sample(l, .1 * len(l))
```

test/perturb.sh
sampsyo, line 1: FWIW, since this is pretty short, it could be inlined in the Makefile. It just needs `\`s at the end of each line to be treated as a single line of shell.

test/turnt.toml
sampsyo: Aha, that is annoying. It would be good if `2>&1` worked in Turnt commands. TBH I am not exactly sure how to accomplish this, but would you mind opening a Turnt bug and we can go from there?

sampsyo (commented):
Just a few low-level suggestions!

susan-garry:
> Re: testing and SE, there is another silly issue: I am occasionally forced to create temp files, and those have to have the .gfa extension in order for odgi to accept them as inputs. Then, because they have the .gfa extension, turnt tries to run tests on these temp files as though they were bona fide test files. Silly and a bit gross, and I'd appreciate any thoughts about cleaning up this workflow.

Maybe the simplest solution is just to create a list of `SLOW_OG_TEST_FILES` and then run `turnt` on this list of actual test files. For example:
```
SLOW_OG_TEST_FILES := $(TEST_FILES:%=test/%.gfa)
// Equivalent to SLOW_OG_TEST_FILES = t.gfa k.gfa note5.gfa ... LPA.gfa

test-slow-validate: fetch
	test/perturb.sh
	-turnt -v --save --env validate_oracle $(SLOW_OG_TEST_FILES)
	turnt -v --env validate_test $(SLOW_OG_TEST_FILES)
```

test/turnt.toml
```diff
@@ -81,14 +81,13 @@ command = "python3 ../slow_odgi/overlap.py {base}.paths < {filename}"
 
 [envs.validate_setup]
 binary = true
-command = "python3 ../slow_odgi/perturb.py < {filename} > temp.{filename}"
+command = "python3 ../slow_odgi/perturb.py < {filename}"
+output.gfa = "-"
```
anshumanmohan, line 85: A silly workaround, but I use the output.extension trick to clobber the GFAs in place. This means that the GFAs need to be re-fetched for future commands, but it does help avoid the situation where we were creating new temp gfas and then seeing turnt use those temp gfas as targets.

test/turnt.toml
anshumanmohan: Done!

slow_odgi/validate.py
```diff
@@ -0,0 +1,29 @@
+import sys
+import mygfa
+import preprocess
+from typing import List, Tuple, Dict
+
+def sego_str(seg):
+    return seg[0] + ("+" if seg[1] else "-")
```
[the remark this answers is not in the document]
anshumanmohan: So I thought this was a neat idea and got to trying it, but then ran into issues in my adjacency matrix (in test/preprocess.py). 

Say I make a new class, SegNameO, in gfa. Now the dict inside the adjmat ends up looking like this:
```
    key: SegNameO              # my details
    value: list of SegNameO    # neighbor's details
```
and this seems to make Python worry about unhashable types being used as dict keys. I figure there is a workaround, but is this getting a bit out of hand? Just wanted to check in before going too deep.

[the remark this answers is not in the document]
anshumanmohan: Ah never mind, I discovered `__hash__`.

anshumanmohan:
Done with the tasks that fell out of your review @sampsyo! The [solution](https://github.com/cucapra/pollen/pull/29#discussion_r1160304303) for perturbing the graphs to generate new .gfas is rather naughty indeed, so please let me know what you think. If you think it's okay (given the alternatives), please hit merge!

slow_odgi/mygfa.py
```diff
@@ -91,19 +91,41 @@ def __str__(self):
             str(self.overlap),
         ])
 
+@dataclass
+class SegO:
```
anshumanmohan, line 95: I realize this looks like a sprawling commit. It's all basically just the claeanup after getting this little dataclass into mygfa.

test/turnt.toml
sampsyo: Something like this makes sense to me! Maybe deserves its own PR though?

slow_odgi/mygfa.py
sampsyo: `SegO` is a fine name, but I think a "segment/orientation pair" is what we also know as a "handle"? (At least according to [the relational doc](https://github.com/cucapra/notes/blob/main/Pollen/backend/relational.md#other-types).) It's a confusing name, obviously, but it is what folks seem to call it.

test/turnt.toml
anshumanmohan: Gotcha, will do!

slow_odgi/mygfa.py
anshumanmohan: 🙈 Wow a little embarrassing! I'll change it!

PR #30, opened 2023-03-21
Slow odgi: `flatten`

anshumanmohan:
This PR adds support for `odgi flatten`. IMO it only really makes sense to run this command with _both_ its flags supplied, though it is happy to run with either. Running
`odgi flatten -i graph.og -f graph.fasta -b graph.bed`
produces:
- in graph.fasta, a simple concatenation of the sequences of the graph's segments. Our visitation (and odgi's) is in the same order that the segments appeared in the original graph.gfa
- in graph.bed, a "key" by which to retrieve path information from the .fasta string. 

It may be clear to see why these really go hand in hand: the former has no path information, and the latter has no sequence information.

It's a simple enough algorithm, and all tests pass with the current suite of graphs.

### DIFF
--- a/slow_odgi/flatten.py
+++ b/slow_odgi/flatten.py
@@ -29,7 +29,7 @@ def insert_newlines(string, every=80):
     return '\n'.join(string[i:i+every] for i in range(0, len(string), every))
 
 if __name__ == "__main__":
-    if len(sys.argv) > 1 and sys.argv[1][-4:] == ".gfa":
+    if len(sys.argv) > 1 and sys.argv[1].endswith(".gfa"):
         infile = open(sys.argv[1], 'r')
         graph = mygfa.Graph.parse(infile)
         odginame = sys.argv[1][:-4] + ".og"
--- a/slow_odgi/flatten.py
+++ b/slow_odgi/flatten.py
@@ -16,7 +16,7 @@ def fasta(graph):
         length = len(segment.seq)
         legend[segment.name] = (ptr, ptr + length)
         ptr += length
-    return(ans, legend)
+    return ans, legend
 
 def print_bed(graph, legend):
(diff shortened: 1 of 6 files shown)


anshumanmohan:
I originally wanted to also do `odgi inject` in the same PR, since it also has to do with fasta and bed. 
In theory, running
`odgi inject -i graph.og -b more_paths.bed -o graph_with_more_paths.og`
produces a new graph, where any paths chalked out in more_paths.bed are inserted. The .bed file needs to be compatible with the given graph.

Sadly I'm not really able to run this as of writing. I made a `k.bed` using `k.og`, then simply changed the name of the path "x" into "z" (i.e., adding a new path that exactly coincides with the known-to-be-correct path "x") but found the resultant graph to be unchanged. I'll investigate further, including snooping around the codebase.

slow_odgi/flatten.py
```diff
@@ -0,0 +1,42 @@
...
+    for path in graph.paths.values():
+        for i, (segname, o) in enumerate(path.segments):
+            start, end = legend[segname]
+            print ("\t".join([odginame, str(start), str(end), path.name, "+" if o else "-", str(i)]))
+
+def insert_newlines(string, every=80):
+    return '\n'.join(string[i:i+every] for i in range(0, len(string), every))
+
+if __name__ == "__main__":
+    if len(sys.argv) > 1 and sys.argv[1][-4:] == ".gfa":
```
sampsyo, line 32:
```suggestion
    if len(sys.argv) > 1 and sys.argv[1].endswith(".gfa"):
```

sampsyo (approved):
Super awesome! I was worried that FASTA files and `.bed` files were going to be more complicated than this. Nice work figuring out these text-based formats to make comparison easy!

PR #31, opened 2023-03-23
Slow odgi: `overlap`

anshumanmohan:
This PR adds support for `odgi overlap`.

The command is straightforward and we diff out cleanly agains the present suite of test graphs. 

One weird thing: the ODGI output is
| #path  | | | path_touched |
| -------------------- | ---------| ----------| ----------------------- |
| name of query path | number1 | number2 |  name of touched path  |
| name of query path | number1 | number2 |  name of touched path  |

but the middle two columns are not labeled, and exploring the [relevant source code](https://github.com/pangenome/odgi/blob/bfae0b3c70a0f044d7ce8510b7a45f59d5138626/src/subcommand/overlap_main.cpp#L167-L168) does not tell me very much. Hilariously, I found, just "by eye", that the second column is always 0 and the third is always the length of the sequence charted by the query path.

I think there is a bug in the odgi code. They want something else, and are trying to compute it, but are inadvertently computing the length of the query path instead. I'll explore further and report back here.

There are some computations in this overlap.py that I'd like to lift into preprocess.py. Also, there is a somewhat ugly bit of .sh hacking in my .turnt file. Thoughts about cleaning that up appreciated!

### DIFF
--- a/slow_odgi/overlap.py
+++ b/slow_odgi/overlap.py
@@ -1,28 +1,19 @@
 import sys
 import mygfa
 
-
-
 def getpaths(infile):
-    paths = []
-    for line in mygfa.nonblanks(infile):
-        paths.append(line)
-    return paths
+    return list(mygfa.nonblanks(infile))
 
 def touches(path1, path2, graph):
     if path1 == path2:
         return False
-    for seg_o1 in graph.paths[path1].segments: # (segment, orientation) tuple
-        for seg_o2 in graph.paths[path2].segments:
-            if seg_o1 == seg_o2:
-                return True
-    return False
+    segs1 = set(graph.paths[path1].segments)
+    segs2 = set(graph.paths[path2].segments)
+    return bool(segs1 & segs2)
 
 def pathseqlen(path, graph):
-    length = 0
-    for seg, _ in graph.paths[path].segments:
-        length += len (graph.segments[seg].seq)
-    return length
+    return sum(len(graph.segments[seg].seq) for (seg, _) in \
+        graph.paths[path].segments)
 
 def print_overlaps(graph, inputpaths):
     print("\t".join(["#path", "start", "end", "path.touched"]))
@@ -30,7 +21,8 @@ def print_overlaps(graph, inputpaths):
         assert (ip in graph.paths)
         for pathname in graph.paths.keys():
             if touches(ip, pathname, graph):
-                print("\t".join([ip, "0", str(pathseqlen(ip, graph)), pathname]))
+                print("\t".join([ip, "0", str(pathseqlen(ip, graph)), \
+                    pathname]))
 
 if __name__ == "__main__":
     if len(sys.argv) > 1 and sys.argv[1][-6:] == ".paths":
--- a/slow_odgi/overlap.py
+++ b/slow_odgi/overlap.py
@@ -25,7 +25,7 @@ def pathseqlen(path, graph):
     return length
 
 def print_overlaps(graph, inputpaths):
-    print("\t".join(["#path","path_touched"]))
+    print("\t".join(["#path", "start", "end", "path.touched"]))
     for ip in inputpaths:
         assert (ip in graph.paths)
         for pathname in graph.paths.keys():
--- a/slow_odgi/overlap.py
+++ b/slow_odgi/overlap.py
@@ -1,0 +1,39 @@
+import sys
+import mygfa
+
+
+
+def getpaths(infile):
+    paths = []
+    for line in mygfa.nonblanks(infile):
+        paths.append(line)
+    return paths
+
+def touches(path1, path2, graph):
+    if path1 == path2:
+        return False
+    for seg_o1 in graph.paths[path1].segments: # (segment, orientation) tuple
+        for seg_o2 in graph.paths[path2].segments:
+            if seg_o1 == seg_o2:
+                return True
+    return False
+
+def pathseqlen(path, graph):
+    length = 0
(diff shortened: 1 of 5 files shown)


test/turnt.toml
```diff
@@ -45,3 +45,15 @@ command = "odgi flip -i {filename} --out=temp.og; odgi view -g -i temp.og"
 [envs.flip_test]
 binary = true
 command = "python3 ../slow_odgi/flip.py < {filename}"
+
+[envs.overlap_setup]
+binary = true
+command = "fn={filename}; pathsname=$(echo $fn | cut -f 1 -d '.').paths; python3 ../slow_odgi/list_paths.py < {filename} > $pathsname"
```
anshumanmohan, line 51: Yuck! 
This is because {filename} has graph.gfa, and I'd like to generate graph.paths. 
Why? Because later, I will have a turnt command (environment "oracle") where {filename} has graph.og, and I want that turnt command to _also_ be able to look for graph.paths and find the graph.paths generated by this setup env.

anshumanmohan:
Meanwhile, a quick heads up for @susan-garry! 

I have learned most of what I know about turnt environments from copying off you, but I wonder if you'd like to take a look at the style I now have going. It may help you avoid 
- the symlink-style(?) fake .og files that live under test/depth
- the per-algorithm turnt.toml files (e.g. test/depth/turnt.toml). 

The main trick I'm leveraging was taught to me by @sampsyo [here](https://github.com/cucapra/pollen/pull/27#issuecomment-1452723038): 
> The way Turnt works is that it looks for turnt.toml by walking "upward" from where the test file itself lives. That means that Turnt works the same way regardless of where you are "standing": running turnt foo.t and turnt ../../../foo.t and turnt bar/baz/qux/foo.t from different working directories (assuming those are all different paths to the same file) are all guaranteed to do the same thing.

sampsyo:
> I think there is a bug in the odgi code. They want something else, and are trying to compute it, but are inadvertently computing the length of the query path instead. I'll explore further and report back here.

Very very interesting! I will say again that this is a _really_ useful outcome of doing some slow reference implementations: we debug the fast implementations (and their documentation). Needless to say, [the docs for `odgi overlap`](https://pangenome.github.io/odgi.github.io/rst/commands/odgi_overlap.html) do not clarify what the columns are supposed to mean; hypothetically, a Python implementation should make that crystal clear.

slow_odgi/overlap.py
```diff
@@ -0,0 +1,39 @@
+import sys
+import mygfa
+
+
+
+def getpaths(infile):
+    paths = []
+    for line in mygfa.nonblanks(infile):
+        paths.append(line)
+    return paths
```
sampsyo, line 10: This is equivalent to:

```py
return list(mygfa.nonblanks(infile))
```

slow_odgi/overlap.py
```diff
@@ -0,0 +1,39 @@
...
+
+def touches(path1, path2, graph):
+    if path1 == path2:
+        return False
+    for seg_o1 in graph.paths[path1].segments: # (segment, orientation) tuple
+        for seg_o2 in graph.paths[path2].segments:
+            if seg_o1 == seg_o2:
+                return True
+    return False
```
sampsyo, line 19: I actually don't know if this is any clearer, but we could do this with set intersection:

```py
segs1 = set(graph.paths[path1].segments)
segs2 = set(graph.paths[path2].segments)
return bool(segs1 & segs2)
```

slow_odgi/overlap.py
```diff
@@ -0,0 +1,39 @@
...
+
+def pathseqlen(path, graph):
+    length = 0
+    for seg, _ in graph.paths[path].segments:
+        length += len (graph.segments[seg].seq)
+    return length
```
sampsyo, line 25: Sorry to be Generator Expression Guy, but:

```py
return sum(len(graph.segments[seg].seq) for (seg, _) in graph.paths[path].segments)
```

test/turnt.toml
sampsyo: Aha! The "setup environment" trick is pretty neat; it is cool that this works.

To avoid the yuckiness about an ignored `.out` file, you can configure Turnt to put the result of this command in a `.paths` file. Just add this:

```toml
output.paths = "-"
```

Then remove the `> $pathsname` redirection, so the stuff goes to stdout. The `"-"` in the above configuration means "get your results from stdout." Turnt will then put that data in `whatever.paths`.

sampsyo (commented):
WAHOO!!! Really really neat that this one turns out to be expressible without too much sadness in Python too. I have some suggestions about the code and also about the Turnt configuration business!

test/turnt.toml
anshumanmohan: Ooh many thanks; I've got this in now. Sadly the following two envs are still ugly, because I need to start with {filename} has looks like graph.gfa (resp., like graph.og) and from each of those envs, reference graph.paths.

anshumanmohan:
Thanks a bunch Adrian, I learned a lot of hacky Python!! The turnt trick is real neat too, though now I have a really silly issue that I'll work on shortly. I need to figure out the bash/turnt incantation that will give me, for example, "q.chop.paths" from "q.chop.gfa" and "q.chop.og". My current solution is very silly and gives me "q.paths" (because I `cut` by "." and grab item #1).

anshumanmohan:
What do you know, it was there all along. I needed the turnt-provided `{base}` instead doing all kinds of magic to `{filename}`.

PR #37, opened 2023-04-06
Slow odgi paths

anshumanmohan:
This PR adds support for a the most basic possible incantation of `odgi paths`. The full command is a monster, with lots of interesting features for investigation and modification, but I realized that I had recreated the command's most basic incantation (i.e., just list the names of the graph's paths in the order that the appear in the GFA) as part of other business anyway.


sampsyo (approved):
Excellent!! Hard to argue with this, given that the functionality was already there. :tada:

PR #40, opened 2023-04-11
Use SegO to define Link

anshumanmohan:
A tiny PR falling out of @sampsyo's observation in https://github.com/cucapra/pollen/pull/39#issue-1661535228 that the newly-defined SegO is, in fact, what Link should be made up of.

### DIFF
--- a/slow_odgi/mygfa.py
+++ b/slow_odgi/mygfa.py
@@ -5,13 +5,12 @@ from enum import Enum
 import re
 
 
-def parse_orient(s) -> bool:
+def parse_orient(o) -> bool:
     """Parse an orientation string as a bool.
-
     In our convention, "True" is forward and "False" is reverse.
     """
-    assert s in ('+', '-')
-    return s == '+'
+    assert o in ('+', '-')
+    return o == '+'
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


slow_odgi/preprocess.py
```diff
@@ -31,7 +31,7 @@ def in_out_edges(graph):
         outs[mygfa.SegO(segment.name, False)] = []
 
     for link in graph.links:
-        ins[mygfa.SegO(link.to, link.to_orient)].append(mygfa.SegO(link.from_, link.from_orient))
-        outs[mygfa.SegO(link.from_, link.from_orient)].append(mygfa.SegO(link.to, link.to_orient))
+        ins[link.to].append(link.from_)
+        outs[link.from_].append(link.to)
```
anshumanmohan, line 35: Nice cleanup because of this change!

slow_odgi/mygfa.py
```diff
@@ -61,54 +60,56 @@ def __str__(self):
...
+    orientation: bool
+
+    @classmethod
+    def parse(cls, s, o) -> "SegO":
+        return SegO(s, parse_orient(o))
+
+    def rev(self) -> "SegO":
+        return SegO(self.name, not self.orientation)
+
+    # We need two str methods because Links and Paths prefer SegOs to be
```
anshumanmohan, line 76: But sadly, also some ugliness.
GFAs format paths like 
`P       x       1+,3+,5+ ...`
but links like
`L       1       +       3       + ...`
and now that the `1+` in the path and the `1       +` in the link are _the same thing_, we need to be able to format the same thing two different ways.

sampsyo: Neato; seems reasonable enough to me!

sampsyo (approved):
Wahoo; this worked great!

PR #43, opened 2023-04-11
use separate turnt outputs to reduce clobbering. closes #34

### LINKED ISSUE #34: Use separate outputs in Turnt test environments
I also mentioned this in https://github.com/cucapra/pollen/pull/27#pullrequestreview-1373400324, but we can simplify our Turnt testing lives a little bit by making the different operators output different files. Here's what things currently look like:
https://github.com/cucapra/pollen/blob/80bee84e31ae15f509a5fd8495b763dec73ddea3/test/turnt.toml#L25-L31

While we want both `depth_oracle` and `depth_test` to work with the same expected-output file, we would like them to use a *different* shared expected-output file from `flip_oracle` and `flip_test` (for example). Turnt's [output configuration](https://github.com/cucapra/turnt#output) lets you tell it what extension to use for the saved stdout. For instance:

```toml
output.depth = "-"
```

…will make it use `whatever.depth` instead of `whatever.out`. This will be nice because it means we don't have to be careful to run all the different operators separately to avoid clobbering. (It may also be possible to make it use `whatever.depth.out` or similar if you prefer that aesthetically.)

This one should probably also wait until after https://github.com/cucapra/pollen/pull/29 and https://github.com/cucapra/pollen/pull/31 land.

anshumanmohan:
A simple PR that helps tidy up our testing a fair bit! As #34 advises, I use turnt's `output.extension` feature to make bespoke .expect files for each algorithm. A much less stressful testing experience!

### DIFF
--- a/test/turnt.toml
+++ b/test/turnt.toml
@@ -1,58 +1,72 @@
 [envs.chop_oracle]
 binary = true
 command = "odgi chop -i {filename} -c 3 -o - | odgi view -g -i - | python3 ../slow_odgi/mygfa.py --nl"
+output.chop = "-"
 
 [envs.chop_test]
 binary = true
 command = "python3 ../slow_odgi/chop.py < {filename}"
+output.chop = "-"
 
 [envs.crush_oracle]
 binary = true
 command = "odgi crush -i {filename} -o - | odgi view -g -i - | python3 ../slow_odgi/mygfa.py"
+output.crush = "-"
 
 [envs.crush_test]
 binary = true
 command = "python3 ../slow_odgi/crush_n.py < {filename}"
+output.crush = "-"
 
 [envs.degree_oracle]
 binary = true
 command = "odgi degree -d --input={filename}"
+output.degree = "-"
 
 [envs.degree_test]
 binary = true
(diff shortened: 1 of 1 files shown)


sampsyo:
Excellent! With this, we're getting closer to being able to just do `turnt -j test/*.gfa` and letting everything run in parallel. There is still the dependency that requires us to do a `--save` run first on the oracle before doing a "check" run on the saved results, but resolving that will take some careful thinking on the Turnt side. Maybe something very roughly in the ballpark of https://github.com/cucapra/turnt/issues/26.

anshumanmohan:
Ah I didn't fully realize that this was the goal! I have written a little more in https://github.com/cucapra/turnt/issues/26#issuecomment-1505738139 but, unless I'm mistaken, my comment there suggests that `turnt -j test/*.gfa` is not going to be possible in our case, right?

sampsyo:
Not for now, at least! I think we should be able to get there eventually, but the problem right now is that we need the "oracles" to go first and the "tests" to go second. Maybe Turnt can help us enforce that itself someday.

PR #48, opened 2023-04-13
Refactor slow-odgi

anshumanmohan:
Nothing too dramatic, just picking up after myself. Even the lowest-level comments are very welcome; I'd love to learn!!

### DIFF
--- a/slow_odgi/mygfa.py
+++ b/slow_odgi/mygfa.py
@@ -73,15 +73,14 @@ class Handle:
     def rev(self) -> "Handle":
         return Handle(self.name, not self.orientation)
 
-    # We need two str methods because Links and Paths prefer Handles to be
-    # string-ified in different formats.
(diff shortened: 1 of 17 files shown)


slow_odgi/mygfa.py
```diff
@@ -73,15 +73,14 @@ def parse(cls, s, o) -> "Handle":
     def rev(self) -> "Handle":
         return Handle(self.name, not self.orientation)
 
-    # We need two str methods because Links and Paths prefer Handles to be
-    # string-ified in different formats.
+    """We need two str methods because Links and Paths
+    have different preferences when converting Handles to string
+    """
```
sampsyo, line 82: Since this isn't "attached" to any specific variable or function, it should probably indeed just be a comment instead of a string literal (docstring).

sampsyo (commented):
Looking good! All great stuff; not much to add here.

PR #49, opened 2023-04-13
add chr6.C4 to test-files

anshumanmohan:
The graph chr6 is not the monster that chr8 is, and should perhaps have been in our fetch-list the whole time. @susan-garry, could you please check if its addition causes any problems on your end? If not, merge away!


susan-garry:
Adding `chr6.C4` to the test suit did not break anything, but it turns out that I broke the `exine` API when I moved the files into a different directory while making the parser. I have fixed it and all of the `exine` and `slow-odgi` tests pass, so I will go ahead and merge this branch

Issue #32, opened 2023-04-05
slow-odgi: points of divergence and confusion


anshumanmohan:
- [x] Match `odgi flip` behavior.
This could come from changes on our end, or by triggering further changes in odgi's code or spec. 
Basically I think there is still a bug in `odgi flip`. Andrea made a fix for me, and this helped in one spot, but introduced an issue elsewhere. See [here](https://github.com/pangenome/odgi/pull/485#issuecomment-1496693803) if curious.

- [ ] Figure out `odgi inject`. 
  - [x] Understand it. 
  I can't recreate the stated behavior of `odgi inject` even for trivial requests. I have asked on Matrix; see [here](https://matrix.to/#/!gfJejgcqPPoQQwBTPY:matrix.org/$QtUIJcktjgfVqVn1_CB8LJOFxtRLMVUhrRbXGh9PcVE?via=matrix.org).
  - [ ] Support it in `slow-odgi`.
- [x] Check with Andrea et al if `odgi overlap` is coded up correctly.
    I _can_ recreate the behavior of `odgi overlap`, but I think they mean to calculate something more subtle and in fact have a bug. I have asked on Matrix; see [here](https://matrix.to/#/!gfJejgcqPPoQQwBTPY:matrix.org/$tngcETjHFpEqLZA9vZS4PcJT77LNjIUWZfpB3eX2tY8?via=matrix.org).
- [x] Figure out what's going on in `odgi groom`. See [this](https://github.com/cucapra/pollen/commit/a2807ebc0f25b8a065ebd8a60311705aed8dd926#r105820995) comment if you're curious. 
  - [x] Make a minimal example.
  - [x] Ask about it on Matrix.

anshumanmohan:
Updates: 
2. Got some clarity re: `odgi inject`. Will just get it done quickly.
3. Also got some clarity re: `odgi overlap`, but this has led to a [new question](https://matrix.to/#/!gfJejgcqPPoQQwBTPY:matrix.org/$VV5L5oW-XjFpL-cEJLSZ31Eiyl_dwAJc-g62f8P7TvI?via=matrix.org): how to issue nontrivial flavors of this command?

sampsyo:
Wow! Thank you for centralizing this "tracking" issue for understanding the bits of odgi we are currently confused about! This will be really useful to have.

This is very superficial and process-related, but one useful thing to do in this kind of GitHub issue is to edit the issue description to have checkboxes, like this:

[...]

Issue #46, opened 2023-04-12
slow-odgi 0.2


anshumanmohan:
This is a parking lot for slow-odgi things that have gotten some attention but are not going to get done in this current push.

- [ ] `odgi groom`. It's quite a beast! See [here](https://matrix.to/#/!gfJejgcqPPoQQwBTPY:matrix.org/$p0NGvZchecT5B_bYhA8UaSTZZlkR8YAgfUGZ8TyP1uU?via=matrix.org) and https://github.com/cucapra/pollen/commit/6c361b2d8fdb29626f2a53bb09d5023d857bd2fa.
- [ ] `odgi flip`. I took a solid crack at it and even found a bug in the odgi implementation, but their fix seems to have created a new point of divergence. I still think our implementation is correct, though of course I'm happy to be proven wrong. Follow the discussion [here](https://matrix.to/#/!gfJejgcqPPoQQwBTPY:matrix.org/$gmFtsxrWSbIQVo_BPWCYclvj5VNBteLoP-dUeremmYo?via=matrix.org).

More to come!

PR #50, opened 2023-04-14
Better testing, more graphs

anshumanmohan:
This PR improves our testing infrastructure a tad. We are (still) doing differential testing against `odgi`, so agreeing with them is always a good sign, but the problem was that the existing graphs were, in many cases, not proving to be very interesting targets for the algorithms. For instance, `crush` was having no effect on any graph, `flip` was only flipping one path in one graph, etc.

I had hoped to generate a bunch of new graphs that would run our algorithms through their paces, but it turned out there wasn't a ton to do. Here's a summary of how the various algorithms are tested after this PR:

1. `chop` takes a command-line argument `n` for the chopping-size to aim for. I test it by generating `n` randomly using a little snippet of Python: `n` is chosen from (2, the length of the graph's longest segment).
2. `crush_n` converts "runs" of `N` into a single character `N`. To test it I have added three new files to test/handmade/.
3. I have added three new files to test `flip`. We already disagreed with the odgi output for the graph note5, and we disagree with them on _all_ the graphs that I have crafted. I will explore this further.
4. I already had a reasonably nice way of testing `validate`: I take a gfa, randomly nix 90% of its Links, and then run `validate` on the newly-crafted gfa.
5. The naive version of `overlap`, where the candidate paths are tested along their entire length, is tested exhaustively at present. The more interesting case, where we query a path along a certain fragment of its length, is not yet implemented in `slow-odgi`. See https://github.com/cucapra/pollen/issues/32#issuecomment-1499839056 for more.
6. IMO `degree`, `depth`,`flatten`, and `matrix`  are tested reasonably well by the existing graphs.


anshumanmohan:
I'll point out a few things that I don't love and would love guidance on.
1. Where to store the handmade files: I currently have them in test/handmade/, and `make clean` does not delete them. 
3. How to run certain algorithms against certain files and still retain sanity: I currently do, for example,
        
        ```
        -turnt --save --env crush_oracle test/handmade/crush*.gfa
        ```

and, in test/handmade, I have `crush1.gfa`, `crush2.gfa`, and `crush3.gfa`. Similar with `flip`. The idea is _not_ to run, say, `flip`, against the graphs that were made especially for `crush`.

Makefile
```diff
@@ -27,6 +27,8 @@ test-slow-chop: og
 	turnt --env chop_test test/*.gfa
 
 test-slow-crush: og
+	-turnt --save --env crush_oracle test/*.og
+	-turnt --env crush_test test/*.gfa
 	-turnt --save --env crush_oracle test/handmade/crush*.gfa
```
anshumanmohan, line 32: Just wanted to provide a pointer to the thing I'm talking about in the comment here: https://github.com/cucapra/pollen/pull/50#issuecomment-1507803443

anshumanmohan:
There was one thing that I was seeking guidance on, but I've moved that into #46 and closed this PR.

susan-garry:
> 1. Where to store the handmade files: I currently have them in test/handmade/, and this means that `make clean` does not delete them. If there's a cleverer/neater way, please let me know!

I can't quite figure out which files you think `make clean` should remove; can't you just add `rm` commands to the `clean` target?

> 2. How to run certain algorithms against certain files and still retain sanity: I currently do, for example, `-turnt --save --env crush_oracle test/handmade/crush*.gfa` and, under test/handmade/, I have `crush1.gfa`, `crush2.gfa`, and `crush3.gfa`. Similar with `flip`. The idea is _not_ to run, say, `flip`, against the graphs that were made especially for `crush`.

I don't think there's anything wrong with running each graph across each algorithm; it doesn't seem to take too much time, and it improves the testing coverage of all of our algorithms. However, if you want to be able to run specific tests on an algorithm for debugging purposes, you could give the `.gfa` files an extra extension, e.g. `.crush.gfa` files are made specifically for `crush`, and `turnt --env crush_test test/handmade/*.crush.gfa` would run tests on just these files.

Makefile
```diff
@@ -86,6 +91,9 @@ clean:
 	rm -rf test/depth/basic/*.out
 	rm -rf test/depth/subset-paths/*.out
 
+	rm -rf test/handmade/*.crush
```
susan-garry, line 94: Just pointing out where `rm` commands have been added to `clean`. Could we do this for the other files that we want `make clean` to remove?

.gitignore
```diff
@@ -12,13 +12,16 @@ test/basic/*.out
 test/subset-paths/*.out
 
 test/temp.*
+test/*.n
 test/*.chop
 test/*.crush
+test/*/*.crush
```
sampsyo, line 19: FWIW, you can gitignore stuff in *any* directory with `**`. Like `test/**/*.crush` to subsume these two lines. Or you can just put a separate `.gitignore` in the `test` directory to avoid repeating `test/` on every line and just put `*.crush` in there to cover all subdirectories.

For details, see: https://git-scm.com/docs/gitignore#_pattern_format

slow_odgi/chop_by.py
```diff
@@ -0,0 +1,17 @@
...
+    Specifically: a random number between 2 and the length of the longest seq.
+    """
+    # random.seed(4)
+    longest_seq = max([len(s.seq) for s in graph.segments.values()])
+    print (random.randint(2,longest_seq))
+
+if __name__ == "__main__":
+    graph = mygfa.Graph.parse(sys.stdin)
+    print_n(graph)
+
```
sampsyo, line 17: This is an interesting strategy! I'm guessing it didn't suffice to hard-code a number that is shorter than most longest seqs, like 4 or something? I just ask because this does add another source of complexity; if the simpler approach will work, maybe we should consider it.

test/turnt.toml
```diff
@@ -1,11 +1,16 @@
+[envs.chop_setup]
+binary = true
+command = "python3 ../slow_odgi/chop_by.py < {filename}"
+output.n = "-"
+
 [envs.chop_oracle]
 binary = true
-command = "odgi chop -i {filename} -c 3 -o - | odgi view -g -i - | python3 ../slow_odgi/mygfa.py --nl"
+command = "odgi chop -i {filename} -c $(cat {base}.n) -o - | odgi view -g -i - | python3 ../slow_odgi/mygfa.py --nl"
```
sampsyo, line 8: It occurs to me that you could avoid the use of the "setup file" here by just invoking the relevant command inline. Something like:

```py
command = "odgi chop -i {filename} -c `python3 ../slow_odgi/chop_by.py < {filename}` -o - | odgi view -g -i - | python3 ../slow_odgi/mygfa.py --nl"
```

That would come at the expense of "re-doing" this calculation every time, but maybe that's not so bad.

sampsyo (commented):
Cool cool! Here are some low-level thoughts about the organization.

> Where to store the handmade files: I currently have them in test/handmade/, and this means that `make clean` does not delete them. If there's a cleverer/neater way, please let me know!

That seems perfectly good to me! You could also (or instead) put the downloaded files in *their* own directory. But this seems perfectly good to me.

> How to run certain algorithms against certain files and still retain sanity: I currently do, for example, `-turnt --save --env crush_oracle test/handmade/crush*.gfa` and, under test/handmade/, I have `crush1.gfa`, `crush2.gfa`, and `crush3.gfa`. Similar with `flip`. The idea is _not_ to run, say, `flip`, against the graphs that were made especially for `crush`.

Also seems perfectly fine. Another alternative would be to have these in their own directory, e.g., `test/crush/1.gfa`, `test/crush/2.gfa`, and so on. But this amounts to more or less the same thing.

slow_odgi/chop_by.py
anshumanmohan: I used to have the hard-coded style (I just used 3) but I developed this little script as part of my fancy new "better testing" push. If 3 is just as good, I'm happy to do that!

.gitignore
anshumanmohan: Many thanks!

Makefile
anshumanmohan: Ah sorry that my original question in https://github.com/cucapra/pollen/pull/50#issuecomment-1507803443 was not very clear. I _want_ `clean` to not delete those gfa files, since they are hand-rolled and don't live anywhere but this repository. So stashing them in handmade/ is my way of shielding them from `make clean`

PR #52, opened 2023-04-18
Fixes in `slow-odgi flip`

anshumanmohan:
There was a little more to `flip` than I had thought: when the algorithm flips a path, it also adds links to the graph in support of the newly-flipped path.  

Splitting hairs a little, there are two options for when such links should be added:
1. _Exclusively_ in support of newly-flipped paths. 
2. In support of _all_ paths, newly-flipped and otherwise.

This is being discussed in https://github.com/pangenome/odgi/issues/496. Arguably this distinction doesn't even matter, since it only really comes up when the input graph is invalid.

Anyway, I currently do (1) and odgi seems to do (2). I am happy to switch to (2); it's just a matter of relaxing the proposition passed to [gen_links](https://github.com/cucapra/pollen/blob/57d9261e2de88535e97351ee08076857515a7636/slow_odgi/flip.py#L65) so it reads `lambda _: True`. The handmade file flip4.gfa shows this divergence. 

We also continue to diverge on the matter of note5.gfa; let's see how that shakes out. Otherwise we diff out correctly versus odgi.

More details in the commit messages if curious!


slow_odgi/flip.py
```diff
@@ -16,21 +17,55 @@ def path_is_rev(path, graph):
...
         for seg in reversed(path.segments):
-            segments.append (mygfa.Handle(seg.name, not seg.orientation))
-        return mygfa.Path(path.name+"_inv", segments, path.overlaps)
+            path_segs.append(mygfa.Handle(seg.name, not seg.orientation))
+        return mygfa.Path(f"{path.name}_inv", path_segs, path.overlaps)
     else:
         return path
 
 
+def dedup(list):
```
sampsyo, line 30: It might be nice to annotate that this only works on lists of links:
```suggestion
def dedup(list: List[mygfa.Link]) -> List[mygfa.Link]:
```

slow_odgi/flip.py
```diff
@@ -16,21 +17,55 @@ def path_is_rev(path, graph):
...
+    new = []
+    for item in list:
+        if item not in new and item.rev() not in new:
+            # odgi seems to consider a link's reverse its own duplicate.
+            new.append(item)
+    return new
+
+
+def gen_links(paths, prop) -> List[mygfa.Link]:
+    """Given a list of paths and a proposition on paths,
```
sampsyo, line 40: I chucked at calling this `pop` (your theorem-prover background taught you well!). No need for any changes, but I might have called this a "predicate" (i.e., a function from something to bool).

sampsyo (commented):
Nice work tracking down this difference! FWIW, my general take here is that it doesn't matter too much if we differ from odgi on invalid graphs. Agreeing in that circumstance seems like would-be-nice territory but not worth prioritizing very high.

You mentioned that we continue to disagree on `note5.gfa`: with apologies for not being able to save state perfectly, can you remind me if that is about the same issue (invalid graphs, where paths don't agree with links) or something else?

anshumanmohan:
### tl;dr: 
1. If we change `odgi flip` as I propose (option 1 above), I believe note5 will also get fixed to my preference: running `flip` on it will become a no-op. 
2. However, it can be argued that the change that `flip` is making on note5 _is_ a no-op, in which case it may be okay to leave as-is. I think this route is stylistically iffy.


### Fuller explanation
Check out [note5](https://github.com/pangenome/odgi/blob/master/test/note5.gfa). I claim this graph does not need to be flipped; the correct response is a no-op. Observe also that this is a valid graph.

In https://github.com/pangenome/odgi/pull/485#issuecomment-1496693803 I show what `odgi flip` does to the graph. The thing is, I think odgi considers those two links to be equal. Slipping briefly into slow-odgi terminology, I think that in general, 
```
Link(
  Handle(from, from_dir),
  Handle(to, to_dir)
)
```
is equal to
```
Link(
  Handle(to, not to_dir)
  Handle(from, not from_dir),
)
```

So really what's happened in the case of note5 is that:
1. The link `L 3 - 4 + 0M` was already in the graph (you can see it on line 11).
2. `flip` scrapped the links and began to build them afresh, with the paths as reference.
3. For some reason `L 4 - 3 + 0M` was proposed, and it was added to the list of links. I'm not sure why it was proposed, since `...,4-,3+,...` does not appear on any path (`...,3-,4+,...` does). 
4. Maybe at some point in the future, `L 3 - 4 + 0M` was proposed but rejected by some kind of `dedup` checker on account of a clash with `L 4 - 3 + 0M`.

If this behavior is okay, I'll move on, I promise! I just think it should be a no-op plain and simple, not a no-op modulo Link-rev-equality.

sampsyo:
Interesting! That equivalence is an interesting observation.

It does occur to me that this may be further invocation that edges should ideally be second-class citizens—mainly reflections of the set of paths, which are what really matter.

PR #61, opened 2023-04-20
Slow odgi doc

anshumanmohan:
The documentation is in! Comments for clarity etc very welcome. @priyasrikumar I wonder if you'd like to give it a look as a sort of warm-up. :)


slow_odgi/README.md
```diff
@@ -0,0 +1,350 @@
...
+1. `flip` disgrees against graphs note5.gfa and flip4.gfa.
+2. `inject` disagrees against graphs DRB1-3123.gfa and chr6.C4.gfa.
+
+
+### Explanation of Commands
+
+The remainder of this document will explain, in some detail, the eleven commands that we have implemented. Below we sometimes elide graph information that is inconsequential to the explanation. Unless specified, this is meant to be read as "don't care" and not as absence. 
+
+
+#### `chop`
```
sampsyo, line 25: What do you think about adding quick, one-sentence explanations here? I guess they would not be too different from what's in the current odgi docs. But starting out with an overview to just say something like this could help people get the idea: "Shortens segments to a given maximum length, while preserving the end-to-end sequence represented by each path." Then we can launch into the example.

slow_odgi/README.md
```diff
@@ -0,0 +1,350 @@
+### Overview
```
sampsyo, line 1: I'd recommend starting with `# slow-odgi` so the document itself has a nice title.

sampsyo (commented):
Looking great; very clear indeed!

PR #76, opened 2023-05-02
Depth of a subset of paths

anshumanmohan:
A minor complication is possible in `depth`: instead of querying the depths of nodes where a crossing by _any_ path counts, we now accept a list of paths that should be included in this depth calculation. This PR beefs up `slow-odgi`'s implementation of `depth` in exactly this way.


anshumanmohan:
The [docs](https://odgi.readthedocs.io/en/latest/rst/commands/odgi_depth.html) explain the flag nicely. We are doing `-s` here, but I don't follow how that differs from `-R`.

PR #80, opened 2023-05-03
Cigar overlaps in paths

anshumanmohan:
At some point I removed our CIGAR-string parsing just started putting `*`. The CIGAR string is optional, and a None is rendered as a `*` in GFA files, so this was relatively okay. However, I worried that there may be cases where odgi was doing something careful with CIGAR strings and we were not.

This PR brings CIGAR-string overlaps back into play. In the algorithms slow-odgi covers thus far, odgi never does anything clever with these strings. It carries them around when the path is unchanged, and drops them on the ground when the paths have changed. We now mimic this behavior in slow-odgi.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,32 +30,38 @@ test-mkjson: og
 #################
 
 # Sets up all the odgi-oracles and then tests slow-odgi against them.
-test-slow-odgi: slow-odgi-all-oracles slow-odgi-all-tests
+test-slow-odgi: slow-odgi-setup slow-odgi-oracles slow-odgi-tests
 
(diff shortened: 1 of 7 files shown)


anshumanmohan:
Whoops, sorry, the previous commit about GFAs instead of OGs when running the oracle of `validate` totally belongs in the basic-bench branch. I was overly hasty in merging that in, and then overly hasty in sticking the change in here 🙈 

This is a band-aid solution; the issue is documented in #81

sampsyo (approved):
Nice; good call! I like that this makes it explicit (in the individual algorithms) where we drop the overlaps intentionally.

PR #82, opened 2023-05-03
Add MIT license to Pollen

anshumanmohan:
I'll point out that this is going into the root directory, not into slow-odgi. If I misunderstood and it should have gone under slow-odgi only, please let me know!


sampsyo:
Let's do the whole thing! Cool!

PR #83, opened 2023-05-03
Monorepo

anshumanmohan:
This PR carves `mygfa`, `slow-odgi`, and `data-gen` into three sibling directories:

```
pollen
|
+-- mygfa/
|   |
|   +-- mygfa/
|   +-- pyproject.toml
|
+-- slow_odgi/
|   |
|   +-- slow_odgi/
|   +-- pyproject.toml
|
+-- data_gen/
|   |
|   +-- data_gen/
|   +-- pyproject.toml
|
+-- test/
|
+-- all/
+-- of/
+-- susans/
+-- stuff/
|
+-- .gitignore
+-- Makefile
+-- LICENSE
+-- README
```

`data_gen` is now its own little command, and can be installed with `flit install --user --symlink` while in `path/to/pollen/data_gen/` . `data_gen` also has a new parser, with `depth` and `simple` as the only subcommands. 

Testing is still via turnt, nothing controversial there, but `simple` is not tested because there is no oracle or basis to test again. I just sorta made it up.

Test with:
1. `make test-slow-odgi`
2. `make test-data-gen`
3. `data_gen simple test/k.gfa`


sampsyo (commented):
Wonderful!! This is very nice and convenient to use!! The top-level organization worked out well.

Here are a couple of suggestions:

* Let's declare a library dependency on `mygfa` to the `pyproject.toml` for `data_gen` and `slow_odgi`. This way, people will get an install-time error when they try to use `slow_odgi`, for example, without `mygfa` installed yet. (I've indicated the syntax for this in one case.)
* It seems like the top-level Pollen README could use a couple of instructions about how to install/use this stuff. As a quick start to install these three tools (sufficient for getting the tests to run), we could recommend `python3 -m pip install --user -e ./mygfa ./data_gen ./slow_odgi`. We could also consider a brief note explaining what `slow_odgi` is and pointing to its own individual README.
* Not very important: what do you think about changing the name of `data_gen` or at least it's executable to be more specific? Could even just be `pollen_data_gen`, if we want to be uncreative.

anshumanmohan:
Thanks Adrian! I've incorporated your suggestions, though now I notice that the top-level README is a little stale. A curious person may go beyond slow-odgi and try to run the `node-depth` accelerator, but be disappointed. I'll check to see if we can patch that up right quick.

PR #84, opened 2023-05-05
Fix inject

anshumanmohan:
I have identified and fixed the issue with `inject`!

We [already know](https://github.com/cucapra/pollen/tree/main/slow_odgi#inject) that `inject`ing one subpath triggers (up to) two tiny `chop`-like operations. These tiny chops happen on (up to) two segments traversed by the original path. After performing these chops, we can:
- Do the straightforward housekeeping to make the existing paths look the same as they did before.
- Insert the new subpath seamlessly.

But what if we need to tiny-`chop` a segment that was itself being traversed in the _reverse_ orientation? Given the graph
```
S	1	ATG
S	2	CCCC
P	x	1-,2+	*
```
and the BED file
```
x	0	1	y
```
the correct output is
```
S       1       AT 	// ?!
S       2       G 	// ?!
S       3       CCCC
P       x       2-,1-,3+        *	// changed in place
P       y       2-      *		// ?!
```
Segment 1 needed to be chopped, but the point at which we chopped segment 1 is perhaps surprising. The link on path `y` is perhaps surprising. The way that path `x` has been fixed up is _not_ surprising if we accept the chop-point of segment 1.

The explanation is this. The original path `x` was traversing segment 1 in the reverse direction, meaning that, when the BED file requested a new path `y` that tracked `x` from index 0 to index 1, the path `y` wanted the character `G` (reading segment 1 _backwards_) and not the character `A` (reading segment 1 forwards).

This problem was only cropping up in the larger graphs because those happen to have more reverse-oriented segment traversals. The issue has nothing to do with flipped _paths_, and nothing to do with graph size.


slow_odgi/slow_odgi/inject.py
```diff
@@ -31,7 +31,10 @@ def where_chop(graph, pathname, index):
             return None
         length = len(graph.segments[handle.name].seq)
         if walk + length > index:
-            return handle.name, index - walk
+            if handle.orientation:
+                return handle.name, index - walk
+            else:
+                return handle.name, length - (index - walk)
```
sampsyo, line 37: Very slick! Almost makes me wish for a little method on `Handle` that does this "direction flipping" itself. Something like:

```py
def pos(self, n):
    """Get the concrete index in the underlying segment sequence corresponding
    to the `n`th nucleotide from the beginning (in the appropriate direction).
    """
    if self.orientation:
        return n
    else:
        return <somehow get the length> - n
```

But of course we don't have access to the length there, so maybe this should actually be a method on `Segment` or something.

sampsyo (approved):
Wahoo!!!! That's a very clear explanation; thanks for the example-driven view, as always.

Random research idea that this makes me think of: what if we could prove that Pollen DSL programs preserve the semantics of variation graphs? That is, some but not all odgi commands are "logical no-ops," in the sense that they may split up segments and rename things and stuff, but the set of nucleotide sequences they represent (i.e., `odgi flatten` would still produce the same flat sequences). Could we somehow prove that these programs are no-ops, so they're free of "bugs" like this one?

slow_odgi/slow_odgi/inject.py
anshumanmohan: Ah this would be real nice, but I think we're a little trapped by our interning goals. Handles don't know about segment lengths, and segments don't know about orientations. And... these are both good things! 

You really need a global view of the graph to be able to do the above. I could lift this little bit of math into `preprocess.py`, if that's any nicer? I could go either way.

sampsyo: No particular need, I say! Just something to keep in mind for the future, IMO, in case it comes up again in a solvable way.

Issue #85, opened 2023-05-08
Proofs about our commands


anshumanmohan:
> What if we could prove that Pollen DSL programs preserve the semantics of variation graphs? That is, some but not all odgi commands are "logical no-ops," in the sense that they may split up segments and rename things and stuff, but the set of nucleotide sequences they represent (i.e., `odgi flatten` would still produce the same flat sequences). Could we somehow prove that these programs are no-ops, so they're free of "bugs"?

_Originally posted by @sampsyo in https://github.com/cucapra/pollen/pull/84#pullrequestreview-1417108141_
  
I love this idea!

I'd love to be able to state and prove, for example:
```
forall g, 
valid g -> 
exists g', flip g g' /\ valid g' /\ meaning g == meaning g'
```
where 
1. `flip a b` is a relation, in this case a function, that says that `b` is the flipped version of `a`. i.e, all the good stuff stated [here](https://github.com/cucapra/pollen/tree/main/slow_odgi#flip).
2. `meaning x` is some beautiful encapsulation of the semantics of a graph `x`, rising above all the kludge of segment-splits and names.

anshumanmohan:
Bit of progress on this, still in `slow_odgi` for now. I have done up `paths_logically_le`. I take a path-forward approach and assert, in the relevant commands (just `chop` and `inject` for now) that the graph we pass is "less than or equal to" the graph we get back. That is, every path that I used to have in the input graph still exists, with the same name and sequence, in the new graph.

This issue talks about logical equality, but I've come up with this logical `<=` approach because I think there's some value there. We can state `==` using `<=` and antisymmetry.

PR #87, opened 2023-05-10
`slow_odgi`: proofs

anshumanmohan:
This PR is a modest step towards #85. 

We'll add further checks as we think of them, but for now I have `paths_logically_le`. I take a path-forward approach and ask, in the relevant commands (just `chop` and `inject` for now) that the graph we pass is "less than or equal to" the graph we get back. That is, every path that I used to have in the input graph still exists, with the same name and sequence, in the new graph.

`flip` does satisfy this condition spiritually, but does not satisfy it as written. The reason is silly: `flip` adds the string "_inv" to the names of the paths that it flips. We do this because odgi does. If we are okay with diverging from odgi in this regard, I can remove this addition of "_inv". 

#85 talks about logical equality, but I've come up with this logical `<=` approach because I think there's some value there. We can state `==` using `<=` and antisymmetry.


anshumanmohan:
Terribly sorry to have added on other PRs onto this one! The failing CI made me panic needlessly and so I decided to fix my mess, but I should have gone about it via a separate PR. 

To clarify, this PR now has three things in it:
1. The proof stuff, as described above.
2. Formatting changes all over `pollen-py/`, to make Black happy.
3. A rename of `pollen-py/` into `pollen_py/`, to my MyPy happy.

The last two are not deep or complicated, but they have big diffs. Thankfully, AFAIK, no-one else is currently working on those files. If you are, and if my changes have made life hard for you, please let me know. I'll revert this stuff and make it right!

anshumanmohan:
I'll note that the "rename of `pollen-py/` into `pollen_py/`" PR also got rid of a log of GFA and temp files that IMO should never have been committed! Please let me know if I made a mistake!

sampsyo:
Super cool! Defining a notion of graph equivalence (or refinement) seems to be the first step toward checking program correctness. That is, we decide something like `paths_logically_le` is the right predicate, we can pick the theorem we want to prove for "no-op" operations g as something like: for `g`, `paths_logically_le(g, f(g))`.

About the names and the `_inv` suffix: it's not clear to me whether we'd want to prove something in the presence of renaming, or change the predicate to ignore names (i.e., to just require that there are the same set of paths in each). Interesting question! But maybe not too hard to manage under either choice.

And no worries about the PR noise; I just looked at the one new `proofs.py` file.

anshumanmohan:
Eyy, we passed MyPy! Merging...

PR #92, opened 2023-05-10
Just some more types

sampsyo:
This was truly unnecessary; it was just a "fun" project for me. It just sprinkles in a bunch more Python type annotations, especially in `mygfa` (which is the easiest). As a result, `mygfa` passes somewhat stricter type enforcement now:

```py
mypy mygfa --disallow-untyped-defs
```

### DIFF
--- a/mygfa/mygfa/mygfa.py
+++ b/mygfa/mygfa/mygfa.py
@@ -13,6 +13,10 @@ def parse_orientation(ori: str) -> bool:
     return ori == "+"
 
 
+LegendType = Dict[str, Tuple[int, int]]
+# A legend is a mapping from segment names to pairs of integers.
+
+
 @dataclass
 class Bed:
     """A BED (Browser Extensible Data) file describes regions of a genome.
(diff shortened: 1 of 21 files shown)


anshumanmohan:
Nice!! I did some word on `pollen_data_gen`, but not all the way. Some of those methods are returning dicts that are quite heterogenous... I don't know if we want to use TypedDicts to get through MyPy, or whether we just want to move on without an annotation.

anshumanmohan:
`mypy slow_odgi --disallow-untyped-defs` now works!

anshumanmohan:
Eh scratch that, I did up `pollen_data_gen` too. I did it with Union types and not TypedDicts. One quibble I have is the type of `**kwargs`. The internet suggested I give it the type `int`, but this doesn't work in our case. I just stuck an `Any` in there at last.

anshumanmohan:
Updated the CI script to run 
```
MYPYPATH=mygfa mypy --disallow-untyped-defs mygfa slow_odgi pollen_data_gen
```
This works correctly on Havarti for me!

anshumanmohan:
☝️ And that was it working correctly on Github Actions. Ready to merge, methinks

PR #101, opened 2023-06-08
Make a new dataclass `Strand`

anshumanmohan:
A segment's seq is now a Strand and not a string. I've done the cleanup required in other parts of the development; mostly just bridging the gaps where a user was quietly using the fact that a segment's seq is a string.

### DIFF
--- a/mygfa/mygfa/mygfa.py
+++ b/mygfa/mygfa/mygfa.py
@@ -44,30 +44,61 @@ class Bed:
 
 
 @dataclass
+class Strand:
+    """A strand is a string that contains only A, T, G, C, or N."""
+
+    strand: str
+
+    def revcomp(self) -> "Strand":
+        """Returns the reverse complement of this strand."""
+        comp = {"A": "T", "C": "G", "G": "C", "T": "A"}
+        return Strand("".join(reversed([comp[c] for c in self.strand])))
+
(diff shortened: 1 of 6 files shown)


mygfa/mygfa/mygfa.py
```diff
@@ -43,31 +43,62 @@ def __str__(self) -> str:
         return "\t".join([self.name, str(self.low), str(self.high), self.new])
 
 
+@dataclass
+class Strand:
+    """A strand is a string that contains only A, T, G, C, or N."""
+
+    strand: str
```
sampsyo, line 50: One other option—and I'm not 100% sure this is a good idea—is to make `Strand` a subclass of `str` rather than a dataclass with one field. That would avoid the need to access `self.strand` in a few places; you could just use `self`. And we wouldn't need `__str__` or `__len__`; those would already be implemented for us.

sampsyo (commented):
Awesome. I love the idea of having a special `Strand` type. See within for one minor suggestion about possibly reducing the verbosity by making this a `str` subclass.

mygfa/mygfa/mygfa.py
anshumanmohan: Oh I like it a lot! Done ~~

PR #102, opened 2023-06-09
Expand `pollen_data_gen simple`

### LINKED ISSUE #96: `pollen_data_gen`: expand `simple`
Presently:
- `pollen_data_gen depth graph.gfa` matches `exine depth -d graph.gfa -a graph.gfa` with some care. The file generated is `graph.data`, which is what Pollen's depth-accelerator needs. `graph.data` contains precomputed stuff beyond the basic graph, but it also lacks some things fundamental to the graph. As a result, this is a lossy conversion: you cannot travel the roundtrip `graph.gfa <-> graph.data`.
- `pollen_data_gen simple graph.gfa` is just something I dreamed up to put the graph into JSON. It has no connection with any accelerator's needs, nor with `graph.data` described above. However, I think it can be shown to be lossless. 

Goals:
- [x] Leave `pollen_data_gen depth` as it is.
- [ ] Change `pollen_data_gen simple`, expanding its precompute and reformatting its output, so that it becomes a superset of `pollen_data_gen depth`. 
- [ ] Show it to be lossless, i.e., capable of a round trip. 
- [ ] Show that it can be filtered in some way to arrive at `pollen_data_gen depth`.

anshumanmohan:
This PR expands the `simple` option of `pollen_data_gen` as sketched out in #96.

That is, `simple` now accepts a graph and emits a lengthy legal JSON that has:
1. No loss, so a roundtrip is possible. 
2. All the info that a `depth` accelerator will want from the graph.

The depth-specific stuff can just be `jq`-ed out of the `simple` output. It matches `exine` exactly, and so this is what we have in turnt now.
Try it with
```
pollen_data_gen simple test/note5.gfa | jq .depth
```

To check that this is lossless, run 
```
pollen_data_gen roundtrip test/note5.gfa
```
Silence is good: it means that an assertion of equality has not been failed.

This code is all kinda idiosyncratic and nasty, and I'd appreciate any thoughts about making it nicer. Docstrings explain the encoding/decoding rules I've made up as I went along.

Closes #96.

### DIFF
--- a/pollen_data_gen/pollen_data_gen/simple.py
+++ b/pollen_data_gen/pollen_data_gen/simple.py
@@ -108,7 +108,7 @@ class GenericSimpleEncoder(JSONEncoder):
         if isinstance(o, mygfa.Link):
             return link_to_number_list(o)
         if isinstance(o, mygfa.Header):
-            return str(o)
+            return o
         if isinstance(o, mygfa.Segment):
             return strand_to_number_list(o.seq)
         return None
--- a/pollen_data_gen/pollen_data_gen/simple.py
+++ b/pollen_data_gen/pollen_data_gen/simple.py
@@ -1,5 +1,4 @@
 import json
-import dataclasses
 from typing import Dict, Union, Optional, Any, List
 from json import JSONEncoder
 from mygfa import mygfa
@@ -47,8 +46,6 @@ class GenericSimpleEncoder(JSONEncoder):
             return str(o)
         if isinstance(o, mygfa.Segment):
             return strand_to_number_list(o.seq)
-        if isinstance(o, (mygfa.Segment, mygfa.Alignment, mygfa.Link)):
-            return dataclasses.asdict(o)
         return None
 
 
--- a/pollen_data_gen/pollen_data_gen/simple.py
+++ b/pollen_data_gen/pollen_data_gen/simple.py
@@ -58,8 +58,8 @@ def dump(graph: mygfa.Graph, json_file: str) -> None:
         json.dump(
             {"headers": graph.headers}
             | {"segments": graph.segments}
-            | {"links": graph.links}
-            | {"paths": graph.paths},
+            | {"paths": graph.paths}
+            | {"links": graph.links},
             file,
             indent=2,
             cls=GenericSimpleEncoder,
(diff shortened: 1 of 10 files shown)


pollen_data_gen/pollen_data_gen/simple.py
```diff
@@ -1,11 +1,97 @@
...
+from typing import Dict, Union, Optional, Any, List, Sequence, TextIO
 from json import JSONEncoder
 from mygfa import mygfa
+from . import depth
 
 
-SimpleType = Optional[Dict[str, Union[bool, str, int]]]
+SimpleType = Optional[
+    Union[str, Dict[str, Sequence[object]], List[int], List[Union[int, str]]]
+]
```
sampsyo, line 11: Perhaps this could use a quick comment about what it's for? I think it may be the general type for all JSON-serializable objects produced here, for instance.

pollen_data_gen/pollen_data_gen/simple.py
```diff
@@ -1,11 +1,97 @@
...
+    ans = []
+    for chunk in path.split(","):
+        num, orient = chunk[:-1], chunk[-1]
+        ans.append(int(num))
+        if orient == "+":
+            ans.append(0)
+        else:
+            ans.append(1)
+
+    return ans
```
sampsyo, line 47: This is not actionable for now, but for future iterations of the hardware, we might want to consider splitting this into two arrays (memories): one for the indices and one for the directions. This would save space because the index memory would be $`\log_2 \text{max\_steps}`$ bits per element and the direction memory would be 1 bit per element.

sampsyo (commented):
Awesome!! Super cool that this can round-trip itself. I have one comment about "de-interleaving" the parts of a path—but I honestly think that could just be left for the future and not addressed here, since it has implications for how we build the hardware that wants to consume this stuff.

One possible agenda item for future synchronous discussions: shall we take this conversion one step farther and produce _binary_ data (rather than just JSON) for fast software processing?

pollen_data_gen/pollen_data_gen/simple.py
anshumanmohan: Ah nice! I've stuck a comment in there for now, so we don't forget.

PR #109, opened 2023-07-08
Use `slow_odgi`'s `paths` subcommand consistently

anshumanmohan:
Until this PR I was using `python -m slow_odgi.paths` in some cases and the actual subcommand `slow_odgi paths` in some. There was some reasoning behind this, and that is chalked out [here](https://github.com/cucapra/pollen/pull/108#issuecomment-1626525948), but it's time to move on and make these uses of the command consistent.


sampsyo (approved):
Excellent; this is helpful!

PR #111, opened 2023-07-08
Simplify and parallelize test execution

sampsyo:
(Depends on, and should be reviewed/merged after, #110.)

This is an initial payoff from the reorganization in the previous few testing-related PRs. Now that I can actually run the tests reliably, I am attempting to simplify the way we run them so it becomes easier to set up CI for them (cf #53). The goal in this PR is to run the test commands in batches and in parallel: that is, instead of doing `turnt -e env1 ; turnt -e env2`, it is possible to do `turnt -j -e env1 -e env2` and run everything in one big parallel batch. This hopefully has the added bonus of making it clearer what orderings are important, which will facilitate the CI automation.

On my machine, the tests run in about 8m40s:

```
$ time make slow-odgi-tests
turnt -j --env chop_test --env crush_test --env degree_test --env depth_test --env flip_test --env flatten_test --env inject_test --env matrix_test --env overlap_test --env paths_test --env validate_test test/*.gfa
1..8
ok 1 - test/DRB1-3123.gfa chop_test
ok 2 - test/DRB1-3123.gfa crush_test
ok 3 - test/DRB1-3123.gfa degree_test
ok 4 - test/DRB1-3123.gfa depth_test
ok 5 - test/DRB1-3123.gfa flatten_test
ok 6 - test/DRB1-3123.gfa flip_test
ok 7 - test/DRB1-3123.gfa inject_test
ok 8 - test/DRB1-3123.gfa matrix_test
ok 9 - test/DRB1-3123.gfa overlap_test
ok 10 - test/DRB1-3123.gfa paths_test
ok 11 - test/DRB1-3123.gfa validate_test
ok 12 - test/LPA.gfa chop_test
ok 13 - test/LPA.gfa crush_test
ok 14 - test/LPA.gfa degree_test
ok 15 - test/LPA.gfa depth_test
ok 16 - test/LPA.gfa flatten_test
ok 17 - test/LPA.gfa flip_test
ok 18 - test/LPA.gfa inject_test
ok 19 - test/LPA.gfa matrix_test
ok 20 - test/LPA.gfa overlap_test
ok 21 - test/LPA.gfa paths_test
ok 22 - test/LPA.gfa validate_test
ok 23 - test/chr6.C4.gfa chop_test
ok 24 - test/chr6.C4.gfa crush_test
ok 25 - test/chr6.C4.gfa degree_test
ok 26 - test/chr6.C4.gfa depth_test
ok 27 - test/chr6.C4.gfa flatten_test
[...]

PR #112, opened 2023-07-08
Use a new Turnt feature to simplify validate tests

sampsyo:
(A successor to #111. Until that's merged, this PR will seem to duplicate those changes.)

We had a problem with the tests for `slow_odgi validate` that made things sorta inconvenient in the Turnt config. The problem was that we needed to produce _invalid_ `.gfa` files, and odgi required them to have that suffix, but this would in turn cause Turnt to pick them up as extra tests.

I added a new feature to Turnt in https://github.com/cucapra/turnt/pull/27 that just lets us quarantine those invalid `.gfa`s in a separate directory so we won't get confused anymore. This makes the validate tests more like other tests we already have and avoids the need for moving around temporary files.


anshumanmohan (approved):
Way cool! It's very nice to see a particularly hacky piece of code go away. So long, 
```for fn in `ls test/*.temp`; do `mv $$fn $${fn%.*}_temp.gfa`; done```!

sampsyo:
Excellent!

Issue #53, opened 2023-04-18
Set up CI


sampsyo:
It would be really great to someday set up CI using [GitHub Actions](https://docs.github.com/en/actions). Ideally, this would let us run our entire test suite on every PR to make sure we don't break anything.

The one tricky thing about this will be getting odgi set up. We could use our existing Dockerfile for that—perhaps more reasonably—we could just run odgi from [the official Docker image](https://github.com/pangenome/odgi#docker). That is, it may be as simple as using `docker run pangenome/odgi` in place of `odgi`, or whatever the Actions-endorsed equivalent is for bringing in a tool from a published Docker image.

anshumanmohan:
Love it; will explore!

anshumanmohan:
I've taken a tiny step towards this by adding a formatting-checker for all our Python code. #87, which presently open, passes this. A more serious CI is coming up!

anshumanmohan:
So I think `odgi` is indeed going to be the big hurdle, and I think using their Docker is not quite going to work. I'd like to show you my thrashing just to make sure I'm correct in dismissing this approach. Please let me know if I missed something!!

```bash
$ # The following won't work
$ docker pull pangenome/odgi
Using default tag: latest
Error response from daemon: manifest for pangenome/odgi:latest not found: manifest unknown: manifest unknown

$ # We actually need to get it from quay, as described here:
$ # https://odgi.readthedocs.io/en/latest/rst/installation.html#docker
$ docker pull quay.io/biocontainers/odgi:0.8.2--py38h3b68952_0
0.8.2--py38h3b68952_0: Pulling from biocontainers/odgi
Digest: sha256:1f77a96eb368331805a01856c0c9432f872033e869b9e2d19de2aa17bc12f760
Status: Image is up to date for quay.io/biocontainers/odgi:0.8.2--py38h3b68952_0
quay.io/biocontainers/odgi:0.8.2--py38h3b68952_0

$ # Now the following should work, but doesn't
$ docker run quay.io/biocontainers/odgi
Unable to find image 'quay.io/biocontainers/odgi:latest' locally
docker: Error response from daemon: manifest for quay.io/biocontainers/odgi:latest not found: manifest unknown: manifest unknown.
See 'docker run --help'.

$ # This works; it needs the tag
$ docker run quay.io/biocontainers/odgi:0.8.2--py38h3b68952_0 odgi
odgi: optimized dynamic genome/graph implementation, version v0.8.2-0-g8715c55

[...]

PR #113, opened 2023-07-09
Collapse make targets for testing

sampsyo:
Now that we've simplified the test setup a bit, it seems straightforward to collapse the remaining "careful" test targets in our `Makefile` into the main ones. Now we just have one set of make targets, organized into this sequence:

1. fetch files
2. setup
3. oracles
4. test

Hopefully this provides a blueprint for the steps we will need for a CI setup (cf #53).

There is a chance I missed an important reason that these targets were separate, in which case @anshumanmohan should please object! :smiley:

---

With all this in place, it becomes obvious that it would be even better if we could run each of these steps with a single `turnt` command (in parallel). Like, all this should be one line:
https://github.com/cucapra/pollen/blob/e10d07fcb301bc3a2506e853e9c77eeda0d119b5/Makefile#L60-L63

The problem is that Turnt does not yet support running different environments for different files. That would be nice to have, but I haven't come up with a design I like for it yet. The solution could be https://github.com/cucapra/turnt/issues/7 or some other way to specify test/environment pairs. But for now, this will do.


Makefile
```diff
@@ -57,23 +59,8 @@ TEST_ENVS := chop_test crush_test degree_test depth_test flip_test \
 slow-odgi-tests:
 	turnt -j $(TEST_ENVS:%=--env %) test/*.gfa
 	turnt -j --env validate_test_err test/invalid/*.gfa
-
-# The basic test suite above, plus a few handmade tests for good measure.
-# Those are described below.
-test-slow-odgi-careful: slow-odgi-crush-careful slow-odgi-flip-careful
```
anshumanmohan, line 63: I had fenced these off as "careful" because they are small handmade tests that are designed to test specific nuances in the odgi specs. Someone could roll their eyes and call them _overly_ careful. I'm all for including them in the main test flow if you are!

sampsyo: Cool; sounds right to me! In a weird way, "testing weird nuances" is kinda the _normal_ role that tests play; our other approach (use a sampling of real-world inputs and hope they cover interesting cases naturally) is less standard. So yeah, I'm all for including our extra-pedantic, targeted tests as part of the normal workflow.

Issue #90, opened 2023-05-10
Redundancies in `pollen-py`?


anshumanmohan:
Running the Black formatter on all our Python files caused me to take an unplanned tour of pollen_py. I'm just wondering if there has maybe been a mistake in committing files after a planned rename/restructure. The structure looks like
```
pollen
|
+- pollen/                   // here be all the new Rust stuff, great!!
+- pollen_py/
     |
     +- depth/
     +- pollen/
          |
          +- depth/
```
and I'm a bit confused. The two `depth/` directories look the same, or at least very similar. Sorry if I'm missing something!

anshumanmohan:
It seems MyPy also thinks this is an issue.
```
mypy $(git ls-files '*.py')
pollen_py/pollen/depth/processing-elements/calyx_depth_simple.py: error: Duplicate module named "calyx_depth_simple" (also at "pollen_py/depth/processing-elements/calyx_depth_simple.py")
pollen_py/pollen/depth/processing-elements/calyx_depth_simple.py: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
pollen_py/pollen/depth/processing-elements/calyx_depth_simple.py: note: Common resolutions include: a) using `--exclude` to avoid checking one of them, b) adding `__init__.py` somewhere, c) using `--explicit-package-bases` or adjusting MYPYPATH
Found 1 error in 1 file (errors prevented further checking)
```

sampsyo:
To crystallize the question for @susan-garry: it looks like the stuff in this directory:
https://github.com/cucapra/pollen/tree/main/pollen_py/pollen

is also duplicated in this directory (alongside project metadata) in this directory:
https://github.com/cucapra/pollen/tree/main/pollen_py

and we should almost certainly delete the latter, keeping the former.

PR #118, opened 2023-07-13
Remove duplicate depth directory in `pollen_py`

susan-garry:
This addresses #90 . `make test` passes for me, but I am having issues with `make slow-odgi-test` even on the main branch, so unless someone else can confirm that the `slow-odgi` tests still pass on their computer, I will wait to merge this PR until I can get `slow-odgi` working!


anshumanmohan:
Hm indeed, something bad has slipped through. It should not be too hard to dive in and check where things went awry. It happened in the last couple days for sure.

anshumanmohan:
I've merged Adrian's fix into main and then merged main into this branch. @susan-garry if all looks well now, please go ahead and merge! 🚀

Issue #114, opened 2023-07-10
Depth instructions don't work as written


anshumanmohan:
I suspect this is something minor, and I'd fix it myself, but I've gotten a little turned around by the way pollen-py is engineered (see also, https://github.com/cucapra/pollen/issues/90).

I get the following error when I run the one-liner test:
```
$ exine depth -a -r test/k.og
[fud] ERROR: `.futil' corresponds to multiple stages: ['futil', 'calyx']. Please provide an explicit stage using --to or --from.
```

And I get the same error when I try the more verbose option:
```
make fetch
make test/k.og
exine depth -o depth.futil
exine depth -d test/k.og -o depth.data
exine depth -r depth.data --accelerator depth.futil
```

I am using the latest copy of Calyx and I have configured it as instructed in our README.

sampsyo:
Ah, this is an artifact of https://github.com/cucapra/calyx/pull/1493. There used to be a stage called `futil`; it is now called `calyx`. Because of a somewhat odd way that fud works w/r/t setting up stages, you may have a `futil` stage lingering on your machine.

Take a look at `fud config`; there is probably a `[stages.futil]` section hanging around. You can edit the file indicated at the top of the `fud config` output and delete that section.

susan-garry:
Yes, this looks like an issue with fud's configuration. You can remove `stages.futil` with `fud config --remove stages.futil`, and you may need to set `stages.calyx.exec` with `fud config -c stages.calyx.exec <path-to-calyx>/.target/debug/calyx`.

anshumanmohan:
Thank you, will check and confirm! But just to be clear, this does not require any changes to the docs at all?

anshumanmohan:
Something is still up. 

I have gotten my `fud check` to look correct:

```
$ fud check
global:
 ✔ /scratch/anshuman/calyx exists.

stages.calyx.exec:
 ✔ /scratch/anshuman/calyx/target/debug/calyx installed.

stages.interpreter.exec:
 ✔ /scratch/anshuman/calyx/target/debug/interp installed.

stages.dahlia.exec:
 ✔ dahlia installed.
 ✔ Found version Dahlia (sha = 188d671, status = clean) (status_is_not dirty).

stages.verilog.exec:
 ✔ verilator installed.
 ✔ Found version 5.006 (>= 5.002).

stages.vcd.exec:
 ✔ vcdump installed.
 ✔ Found version 0.1.2 (>= 0.1.2).

stages.jq.exec:
[...]

Issue #98, opened 2023-05-24
Error when generating a node depth accelerator using `exine depth`


tonyjie:
Hi, I'm trying to run some Pollen before our meeting. I basically followed the installation guide for Pollen, Calyx, ODGI, and got an error when generating a node accelerator [here](https://github.com/cucapra/pollen#generating-an-accelerator-quick). I think the error is still something related to ODGI-python binding. 

Do you have any idea on how to fix this?

## Error message
```
(py39) ➜  ~/Other_Repos/pollen/test/depth git:(main) exine depth -a -r t.og
warning [libhandlegraph]: Serialized object does not appear to match deserialzation type.
warning [libhandlegraph]: It is either an old version or in the wrong format.
warning [libhandlegraph]: Attempting to load it anyway. Future releases will reject it!
Traceback (most recent call last):
  File "/home/jl4257/.local/bin/exine", line 8, in <module>
    sys.exit(main())
  File "/home/jl4257/.local/lib/python3.9/site-packages/pollen/main.py", line 26, in main
    depth.run(args)
  File "/home/jl4257/.local/lib/python3.9/site-packages/pollen/depth/main.py", line 180, in run
    run_accel(args, tmp_dir_name)
  File "/home/jl4257/.local/lib/python3.9/site-packages/pollen/depth/main.py", line 104, in run_accel
    parse_data.run(args)
  File "/home/jl4257/.local/lib/python3.9/site-packages/pollen/depth/parse_data.py", line 281, in run
    max_nodes, max_steps, max_paths = get_dimensions(args)
  File "/home/jl4257/.local/lib/python3.9/site-packages/pollen/depth/parse_data.py", line 190, in get_dimensions
    max_nodes, max_steps, max_paths = get_maxes(filename)
  File "/home/jl4257/.local/lib/python3.9/site-packages/pollen/depth/parse_data.py", line 167, in get_maxes
    graph.load(filename)
RuntimeError: Error rewinding to load non-magic-prefixed SerializableHandleGraph
```

## My current environment
- python 3.9.16
- `import odgi` is successfull after I followed the detailed installation [here](https://github.com/cucapra/pollen#odgi) from bioconda. 
- Calyx installed. `fud check` shows the following messages. 
```
global:
 ✔ /home/jl4257/Other_Repos/calyx exists.

stages.calyx.exec:
 ✔ ./target/debug/calyx installed.
   ./target/debug/calyx is a relative path and will not work from every directory.

stages.interpreter.exec:
[...]

PR #120, opened 2023-07-13
Update README.md

susan-garry:
This addresses #114 (it turns out that our calyx installation instructions did need to be updated as a result of calyx renaming `futil` to `calyx` - sorry for the confusion Anshuman!) and the last two comments in #98.

What this PR does:
- Clarify that our installation instructions assume that the user has installed calyx from source
- Add a line about installing the calyx interpreter
- Instruct users to configure `stages.calyx.exec`, rather than `stages.futil.exec`
- Update instructions regarding the version of turnt that we're using

What this PR does NOT do:
- Address the symlinked test files
- Add a more precise invocation of `fud check` to only look at the tools a pollen user would need to get started
- Address how to install and configure calyx via Docker or as a crate

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ and run `cd pollen_py && flit install -s --user`. You will need [`flit`][flit]. 
 
 #### Calyx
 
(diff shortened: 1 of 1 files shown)


anshumanmohan:
Thanks, looks good!

sampsyo (commented):
Ah, nice catch on the erroneous `fud config` command!! Sorry for not noticing this, @anshumanmohan; I mis-diagnosed your issue. Looks good, modulo a capitalization suggestion.

anshumanmohan:
This PR will not close #114. More details [here](https://github.com/cucapra/pollen/issues/114#issuecomment-1634996009). For now I am going to un-link this from 114 and merge it.

Issue #103, opened 2023-06-09
Update installation instructions


anshumanmohan:
Because of changes in [the Calyx docs](https://docs.calyxir.org/intro.html), our own installation instructions are becoming harder and harder to follow. For instance, we point users to the "first and third" steps of the Calyx installation instructions, but the first step now itself has three options (docker, crate, source).

I'm not quite sure what a permanent fix would be, but let's at least patch it some. This can be fixed in the same spirit, if not the same PR, as the fix for #98. 

**First: Crate good?**
@susan-garry, do you know if the crate version of Calyx is good enough for our purposes? That would give us a certain barrier of stability I suppose, since we'd be asking people to go through the crate, not having them pull the bleeding-edge main branch every time. I don't know if that's actually a fair assessment; maybe the Calyx crate is deployed automatically and often.

**Second: This much madness is too much sorrow.**
1. I wonder if we should just inline our own installation instructions for Calyx. I don't love that we currently have users jumping to the Calyx docs, running/skipping stuff there, coming back to us, running more `fud config` stuff, and then getting going. If we inline Calyx installation instructions that we know will work for Pollen's purposes, that will introduce a maintenance task.
2. Regardless of what we do with the inlined/foreign instructions, I'd love to finishw with a `fud check` invocation that passes 100%, with no warnings for the purposes of Pollen. We have [recently learned](https://github.com/cucapra/calyx/issues/1551#issuecomment-1583700031) the format for doing that.

**Third: Docker all the way?**
Perhaps somewhat at odds with all of the above, but this is all making me wonder if we shouldn't just make Docker the default way we do business. Our Dockerfile currently starts with Calyx's Docker image, installs odgi from source, and installs Pollen. 
1. We could catch it up to install mygfa, slow-odgi, and pollen-data-gen, and then legitimately do all our development in Docker. 
2. If we wanted, we could use Adrian's Depot trick to keep the Docker current. Perhaps overkill for a small project like ours.
3. When it comes to coding day to day, VS Code has support for "attaching" to a Docker container, which is the same as SSH-ing into a remote machine like Havarti. 
4. Sorry if I'm forgetting something that we've already resolved, but does this not work for some reason? Is a Docker too underpowered? Does the Calyx Docker lack an important simulation tool that Havarti has? Paging @rachitnigam on this one.
[...]

PR #122, opened 2023-07-14
Straight-line Calyx installation instructions

anshumanmohan:
This is a step towards #103. Instead of having folks visit Calyx's docs and jump around between their docs and our docs, I provide a straight-line recipe. It's nine steps, which is annoying, but ah well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,24 +39,22 @@ and run `cd pollen_py && flit install -s --user`. You will need [`flit`][flit]. 
 
 #### Calyx
 
-Follow these [instructions](https://docs.calyxir.org/) to install Calyx. You must complete the [first](https://docs.calyxir.org/#installing-from-source-to-use-and-extend-calyx) and [third](https://docs.calyxir.org/#installing-the-command-line-driver) sections, installing Calyx from source, but feel free to skip the second section. Then, from the root of the Calyx repository, install the Calyx interpreter by running 
-```
-cargo build -p interp
-```
+Below we show you how to build Calyx from source and set it up for our use.
+We suggest you make `calyx` and `pollen` sibling directories.
 
-The last step should be running `fud check`, which will report that some tools are unavailable. This is okay for our purposes.
+1. `git clone https://github.com/cucapra/calyx.git`
+2. `cd calyx`
+3. `cargo build`.
+3. `flit -f fud/pyproject.toml install -s --deps production`
+4. `fud config --create global.root <full path to Calyx repository>`
+5. `cargo build -p interp`
+6. `fud config stages.calyx.exec <full path to Calyx repository>/target/debug/calyx`
+7. `fud config stages.interpreter.exec <full path to Calyx repository>/target/debug/interp`
+8. `cd calyx-py; flit install -s; cd -`
+9. `fud check`
 
-After completing the above, run
-```
-fud config stages.calyx.exec <full path to calyx repository>/target/debug/calyx
-fud config stages.interpreter.exec <full path to calyx repository>/target/debug/interp
-fud check
-```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,24 +39,22 @@ and run `cd pollen_py && flit install -s --user`. You will need [`flit`][flit].
...
-fud config stages.calyx.exec <full path to calyx repository>/target/debug/calyx
-fud config stages.interpreter.exec <full path to calyx repository>/target/debug/interp
-fud check
-```
+Below we show you how to build Calyx from source and set it up for our use.
+We suggest you make `calyx` and `pollen` sibling directories.
+
+1. `git clone https://github.com/cucapra/calyx.git`
+2. `cd calyx`
+3. `cargo build`.
```
sampsyo, line 47:
```suggestion
3. `cargo build`
```

README.md
```diff
@@ -39,24 +39,22 @@ and run `cd pollen_py && flit install -s --user`. You will need [`flit`][flit].
...
-The last step should be running `fud check`, which will report that some tools are unavailable. This is okay for our purposes.
-
-After completing the above, run
-```
-fud config stages.calyx.exec <full path to calyx repository>/target/debug/calyx
-fud config stages.interpreter.exec <full path to calyx repository>/target/debug/interp
-fud check
-```
+Below we show you how to build Calyx from source and set it up for our use.
+We suggest you make `calyx` and `pollen` sibling directories.
```
sampsyo, line 43: Maybe we can preserve the links to the Calyx docs here too, just so someone who wants more details can find them?

README.md
```diff
@@ -39,24 +39,22 @@ and run `cd pollen_py && flit install -s --user`. You will need [`flit`][flit].
...
+
+1. `git clone https://github.com/cucapra/calyx.git`
+2. `cd calyx`
+3. `cargo build`.
+3. `flit -f fud/pyproject.toml install -s --deps production`
+4. `fud config --create global.root <full path to Calyx repository>`
```
sampsyo, line 49:
```suggestion
4. `fud config --create global.root $(pwd)`
```

Since they're standing in the Calyx repo already, this is an easy way to get that directory's path.

README.md
```diff
@@ -39,24 +39,22 @@ and run `cd pollen_py && flit install -s --user`. You will need [`flit`][flit].
...
+5. `cargo build -p interp`
+6. `fud config stages.calyx.exec <full path to Calyx repository>/target/debug/calyx`
+7. `fud config stages.interpreter.exec <full path to Calyx repository>/target/debug/interp`
```
sampsyo, line 52: Are these necessary? I thought they got configured this way automatically.

(If they are, then the `$(pwd)` trick could work here again.)

README.md
```diff
@@ -39,24 +39,22 @@ and run `cd pollen_py && flit install -s --user`. You will need [`flit`][flit].
...
+8. `cd calyx-py; flit install -s; cd -`
```
sampsyo, line 53: It's possible to use the `flit -f` trick again to make this simpler:
```suggestion
8. `flit install -f calyx-py/pyproject.toml -s`
```

sampsyo (commented):
Nice; this seems like a good idea! Here are a few assorted comments.

README.md
anshumanmohan: Ah, so I had the same thought! I tried `fud check` on a fresh copy of Calyx just before running these two steps. The stages exist, but are accompanied by a warning that they aren't globally available and won't work from any which directory. Those warnings go away after these two steps.

sampsyo: OK, sounds good! It would be nice for fud to figure this out itself someday, since the `global.root` config should theoretically suffice.

PR #123, opened 2023-07-15
Small `slow-odgi` testing fixes, hopefully to fix #119

### LINKED ISSUE #119: Some `slow-odgi` tests failing
Probably minor, but something bad has slipped through. It should not be too hard to dive in and check where things went awry. It happened in the last couple days for sure.

_Originally posted by @anshumanmohan in https://github.com/cucapra/pollen/issues/118#issuecomment-1634672473_

sampsyo:
This fixes a couple of issues in the `slow-odgi` testing setup pointed out in #119. With this, on my machine, cleaning everything and running it from scratch produces only two failures:

```
not ok 17 - test/note5.gfa flip_test # differing: test/note5.flip
not ok 4 - test/handmade/flip4.gfa flip_test # differing: test/handmade/flip4.flip
```

If this seems to work good for one of @susan-garry & @anshumanmohan, these should at least get us closer to everything working. It would be great to know from @susan-garry whether there are any more failures remaining after this (and if so, which).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ TEST_ENVS := chop_test crush_test degree_test depth_test flip_test \
 	 flatten_test inject_test matrix_test overlap_test paths_test validate_test
 slow-odgi-tests:
 	-turnt -j $(TEST_ENVS:%=--env %) test/*.gfa
-	-turnt -j --env validate_test_err test/invalid/*.gfa
+	-turnt -j --env validate_test test/invalid/*.gfa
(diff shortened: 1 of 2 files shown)


anshumanmohan:
Can confirm, same on my end!

susan-garry:
Same for me! I know that `note5.flip` is expected to fail; should we expect `handmade/flip4.flip` to fail as well?

anshumanmohan:
Yes, this is a known point of divergence. https://github.com/pangenome/odgi/issues/496

PR #125, opened 2023-07-19
More complete Dockerfile

anshumanmohan:
The Dockerfile now installs `pollen-py`, `mygfa`, and `slow-odgi`. I have also built and published this package to the repository.


sampsyo (approved):
All looks good!!

PR #126, opened 2023-07-19
Make README point to Docker as the first resource

anshumanmohan:
The README now points to the Docker package as the first resource. The old instructions are still available in their own section at the bottom of the file.

This goes some way towards closing #103, but probably not all the way: see #127.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,15 @@ We are designing a graph-manipulating DSL that exposes functionality that pangen
 Our DSL will support graph queries in the vein of the [odgi][] project.
 We will compile programs written in this DSL into the [Calyx][] IR and then leverage Calyx to generate hardware accelerators.
 
+Installation using Docker
+-------------------------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,70 +11,28 @@ We are designing a graph-manipulating DSL that exposes functionality that pangen
 Our DSL will support graph queries in the vein of the [odgi][] project.
 We will compile programs written in this DSL into the [Calyx][] IR and then leverage Calyx to generate hardware accelerators.
 
+Installation using Docker
+-------------------------
+Installation is easy if you use our Docker [package][]:
```
sampsyo, line 16: I know this is a ridiculous nitpick, but I'd shy away from calling the Docker path "installation"… we're not really installing anything anywhere; just running a container where everything is pre-installed. So maybe we should just call this section "Running using Docker" or similar?

sampsyo (commented):
Great! Here are just a couple of small suggestions.

PR #131, opened 2023-07-20
Ask users to install odgi from source

### LINKED ISSUE #103: Update installation instructions

### LINKED ISSUE #127: Most foolproof way to install odgi?
When doing up #126 I was reminded that the installations we provide for installing odgi are wanting. See [this](https://github.com/cucapra/pollen/tree/better-readme#odgi). We sorta point them to too many places: bioconda, installing from source (which has a sorta vague pointer to path-editing), and we finally walk through a hacky middle-ground option.

What do people actually use? I use a built-from-source copy, and I've done the path-modification thing to get Python bindings to work:
```
export PYTHONPATH=/scratch/anshuman/odgi-py/lib/python3.9/site-packages/
export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libjemalloc.so.2
```
It's not that bad. Does any of us do differently? Thoughts? These instructions will now live in the bottom of the file and will be intended for the most serious users, those who want not just to use Pollen but to develop and extend it. I think we can give them one straight-line route, even if it's a little involved.

anshumanmohan:
Tells users to install odgi from source. Closes #127. Closes #103.


sampsyo (approved):
Yay; looks great!

PR #132, opened 2023-07-20
Tidy `parse_data`

anshumanmohan:
I had done some tidying in `parse_data` as part of #116, but I then reverted those because that PR was no longer making any meaningful changes to that file. This quick RP has _just_ the tidying. It should have no meaningful impact anywhere.


sampsyo (approved):
This all looks great! Thanks for separating this out.

PR #134, opened 2023-07-26
Updated Pollen examples!

priyasrikumar:
Currently, no programs in this file parse successfully when run with `cargo run test.pol`. However, I believe I've updated the syntax for existing programs in line with what @susan-garry has been envisioning. 

It could be good to make each of these programs their own file in the `test` directory, if they look good to folks! I kind of like the `.pol` extension name but totally open to suggestions, haha. 

I would love to hear thoughts on any syntactical differences and where these example programs should ultimately live!


slow_odgi/examples.md
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    type Step = {
+      path: Path; 
+      idx: int; //where in the path is this sequence?
+      handle: Handle; //segment + orientation
+    }
+
+    type Base = A | C | T | G | N
+    type Strand = list Base
```
sampsyo, line 33: These ones look especially pseudocode-y, FWIW!

slow_odgi/examples.md
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+## Chop
+
+## Crush
+
+    graph out_g;
+    parset out_segs[Segment, out_g];
+
+    for segment in Segments {
```
sampsyo, line 47: Should `Segments` be `in_g.segments`?

slow_odgi/examples.md
```diff
@@ -0,0 +1,157 @@
...
+      start: Handle; 
+      end: Handle;
+    }
+
+## Chop
+
+## Crush
+
+    graph out_g;
+    parset out_segs[Segment, out_g];
```
sampsyo, line 45: I kinda forget what the two arguments in the square brackets here are supposed to convey. Is the second one always a graph, for example?

slow_odgi/examples.md
```diff
@@ -0,0 +1,157 @@
...
+            seq.push(c);
+        }
+      }
+      emit { segment with sequence: seq } to out_segs;
+    }
+
+## Degree
+
+    // Output is a (Segment, int) array.
+    outset out_degs[Segment, int];
```
sampsyo, line 67: Is `outset` different from `parset`, or is this perhaps a typo?

slow_odgi/examples.md
```diff
@@ -0,0 +1,157 @@
...
+            in_n = false;
+            seq.push(c);
+        }
+      }
+      emit { segment with sequence: seq } to out_segs;
+    }
+
+## Degree
+
+    // Output is a (Segment, int) array.
```
sampsyo, line 66: I now this is just a comment, but it would probably be a little clearer to say we're emitting a set, not an array (because the order doesn't matter).

slow_odgi/examples.md
sampsyo, line 67: Further questions about what the stuff in the square brackets mean. Does this mean that the output set contains `(Segment, int)` tuples? If so, it seems like the more obvious syntax would be `out_degs[(Segment, int)]`. Otherwise, something like `out_stuff[Segment]` would seem to contain 1-element tuples, which doesn't seem right.

slow_odgi/examples.md
```diff
@@ -0,0 +1,157 @@
...
+## Flip
+
+    graph out_g;
+    parset out_steps[Step, out_g];
+    parset out_paths[Path, out_g];
+
+    local is_rev(Path path) {
+      fw = 0;
+      bw = 0;
+      for step.handle as sh in path.steps {
```
sampsyo, line 93: I'm a little confused about this `for x as y in z` syntax. What is `x` allowed to be—for example, in this situation, what does `step` refer to? I think this stuff may be equivalent to:

```
for step in path.steps {
  sh = step.handle;
  ...
}
```

Is that right? If so, I would love to stick to that simpler use of `for`. I think putting a "dot expression" like `step.handle` in the "binding" position of the loop raises a lot of questions about how we want the program to behave.

slow_odgi/examples.md
```diff
@@ -0,0 +1,157 @@
+# Pollen language walkthrough
+
+## Core types in Pollen
```
sampsyo, line 3: One thing I don't quite understand: is this meant to be valid Pollen source code? (And if so, is it supposed to be part of a "preamble" that will be prepended to every program?) Or is it just pseudocode, not actually meant to be parsed, but just to give humans an idea of what types are available to them?

The latter is straightforward enough, but if it's the former, there are some questions here about what these declarations are supposed to convey to the compiler. Do they describe how to load in the data, for example? If this is just meant as a human hint, maybe a brief note to that effect here would be good.

sampsyo (commented):
Thanks for getting this started!

I have a few syntax questions that could turn into longer discussions. Meanwhile, however, yes—please do reorganize these into separate `.pol` files that we can feed into the parser. (This stuff should probably also not go in the `slow_odgi` directory, which is for something else.) Then, could you (hopefully with @susan-garry's help) make a list of tasks in an issue cataloging everything that cannot be parsed—indicating that either the program or the parser must change?

slow_odgi/examples.md
priyasrikumar: Yes, good catch!

slow_odgi/examples.md
priyasrikumar: Good point! Thanks!!

slow_odgi/examples.md
priyasrikumar: Yeah, definitely wanted to rework that because it felt clunky. Thanks for the suggestion!

slow_odgi/examples.md
priyasrikumar: Noted, definitely took streamlining the code a bit far there and obfuscated things. Will revert to your suggestion above!

slow_odgi/examples.md
priyasrikumar: I meant it as the latter! I'll make a note of that, and it'll live apart from the `.pol` test cases, so that it's very clear what it is.

slow_odgi/examples.md
priyasrikumar: I meant to discuss this at the next meeting-- `outset` is the keyword I made up for when we're not modifying a graph, but outputting information about it (for instance, degree/depth). Not a huge fan of the word itself so totally open to changes!

slow_odgi/examples.md
priyasrikumar: This is based off of syntax I got from @susan-garry, but I think the first argument is the graph set that is being modified (i.e., `Segment`s or `Path`s and the second is the graph to which that set belongs.

sampsyo:
Thanks for splitting things up!! Maybe a good next step would be to merge this PR as-is and make a little to-do list covering the ways we want to update the syntax (and questions we still have about how to do that)?

PR #138, opened 2023-09-18
Pollen supports the array access postfix operator

susan-garry:
Pollen can now parse array access statements with arbitrary index expressions, such as `path.steps[path.steps.size() - 1]`!

This means that `paths.pollen` parses correctly, so now all of the `slow-odgi` examples that Priya translated to pollen parse correctly.


sampsyo:
Yay; awesome that this worked out!

PR #140, opened 2023-09-18
Pollen supports array access

susan-garry:
Pollen can now parse array access statements with arbitrary index expressions, such as path.steps[path.steps.size() - 1]!

This means that paths.pollen parses correctly, so now all of the slow-odgi examples that Priya translated to pollen parse correctly.

### DIFF
--- a/pollen/src/pollen.pest
+++ b/pollen/src/pollen.pest
@@ -115,8 +115,8 @@ obj_initialize = { typ ~ "()" }
 call_args = { "(" ~ (expr ~ ("," ~ expr)*)? ~ ")" }
 func_call = { identifier ~ call_args }
 
-term = _{ (literal | obj_initialize | func_call | identifier 
-          | "(" ~ expr ~ ")" | "[" ~ expr ~ "]") }
+term = _{ literal | obj_initialize | func_call | identifier 
+          | "(" ~ expr ~ ")" | "[" ~ expr ~ "]" }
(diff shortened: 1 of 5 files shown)


pollen/src/pollen.pest
```diff
@@ -118,7 +118,10 @@ func_call = { identifier ~ call_args }
 term = _{ literal | obj_initialize | func_call | identifier 
           | "(" ~ expr ~ ")" | "[" ~ expr ~ "]" }
 
-expr = { prefix* ~ term ~ (binop ~ prefix* ~ term ) * }
+array_access = { "[" ~ expr ~ "]" }
+postfix = _{ array_access }
+
+expr = { prefix* ~ term ~ postfix* ~ (binop ~ prefix* ~ term ~ postfix*) * }
```
sampsyo, line 124: This is beyond the scope of this PR, but I could imagine making a few intermediate nonterminals here to make this more intelligible. Something like this:

```
term_ext = { prefix* ~ term ~ postfix* }
binop_chain = { term_ext ~ (binop ~ term_ext)* }
```

sampsyo (approved):
Great!!! Awesome that this works now!

pollen/src/pollen.pest
susan-garry: Looks great! I went ahead and changed that as part of this PR

PR #142, opened 2023-09-24
Try setting up CI

sampsyo:
A totally not ready, experimental approach to #53. Nothing to see here yet.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,14 +53,11 @@ jobs:
       - name: Install odgi alias
         run: |
           echo '#!/bin/sh\ndocker run --rm odgi odgi $@' > odgi
-          mkdir -p $HOME/.local/bin
           cp odgi $HOME/.local/bin/odgi
-          sudo chmod a+x $HOME/.local/bin/odgi
-          echo $HOME/.local/bin/odgi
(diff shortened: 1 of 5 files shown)


sampsyo:
One note as this starts to take shape: so far, this only runs the tests for slow-odgi. It should eventually also run the tests for the data generator and the new tests for the parser added in #136.

.github/workflows/build.yml
```diff
@@ -0,0 +1,76 @@
...
+
+on:
+  push:
+    branches:
+     - main
+  pull_request:
+    branches:
+     - main
+
+jobs:
```
anshumanmohan, line 11: Nice, this feels a lot like our Dockerfile. Which is a good thing. In fact maybe in the future the Dockerfile should learn from your system of sourcing and running odgi.

sampsyo: Yeah, totally. I stole some of the stuff directly from the Dockerfile. Someday merging these things could be cool, but I have never really figured out the way to do this in general for other projects (i.e., to use a "base" Dockerfile twice, once for the "real" Dockerfile and once for the CI).

sampsyo:
Excellent; merging this now!

PR #145, opened 2023-10-30
restore handmade tests to their state immediately before the parser…

susan-garry:
…merge, at commit #134


sampsyo:
Excellent; thanks for fixing this!

PR #149, opened 2023-11-17
Slow odgi depth no paths

sampsyo:
For no particular reason, I noticed that `slow_odgi depth` was requiring you to specify a subset of the paths. The real `odgi` command doesn't require a `--paths` list, so let's not require one either.


anshumanmohan:
Ah, many thanks!! I probably added the paths thing as an extra feature, but then forgot to make it optional as it is in `odgi`.

anshumanmohan (approved):
Thank you! Please feel free to merge!

PR #169, opened 2024-04-27
Tidy `paths` command for blogpost

anshumanmohan:
The code for the command `paths` was a little muddied because I had a whole little option for a "droprate", meaning that a certain percentage of the graph's paths would be dropped from the output. This is not a flag that you'd realistically want to pass to the `paths` command in odgi, but rather a convenience for us: it allowed us to generate a subset of a graph's paths and then pass that subset along when testing the commands `depth` and `overlap`.

This PR separates the command and the utility into their own files. `slow_odgi paths` now takes no options, and is backed up by a short bit of code. A new command `slow_odgi somepaths` now exists, and that is what we use as a setup utility for testing `depth` and `overlap`.


sampsyo:
Awesome!! Good thinking.

PR #171, opened 2024-05-09
Proof-of-concept Python bindings for FlatGFA

sampsyo:
I was inspired to try seeing how hard it would be to build Python bindings for the FlatGFA library. It worked out OK, actually! Check out the README for how to use it, and `example.py` for what the interface looks like. It does almost nothing currently—you can print out info about segments and not much more. Proof of concept!

I got a lot of mileage out of [Maturin][] and [PyO3][]. I'm actually not entirely sold on PyO3, however… it seems amazing for cases where you need very deep interaction with the Python interpreter, but for simply exposing an API it seems like overkill. It introduces a lot of challenging lifetime questions that do not seem essential for simple use cases like this. So I think I'd like to try switching to a more straightforward cffi-based library in the future before going farther.

[maturin]: https://www.maturin.rs
[pyo3]: https://pyo3.rs/


flatgfa-py/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+#[pyclass]
+struct SegmentList {
+    gfa: GFARef,
+}
+
+#[pymethods]
+impl SegmentList {
```
rachitnigam, line 57: Out of curiosity: are these methods generated automatically or you wrote them down by hand? If its the latter, then how do you figure out if they're interacting safely with the Python API?

PR #172, opened 2024-05-09
Set up CI for basic FlatGFA tests

sampsyo:
Currently, we just do the round-trip GFA parser tests. We should someday set up differential testing against odgi/slow-odgi...

This PR will need some CI debugging, certainly, before it's ready to merge.


sampsyo:
Wow, that was surprisingly painless… it works!

PR #180, opened 2024-05-15
Try setting up CI for Python bindings tests

sampsyo:
Fortunately, [Maturin](https://www.maturin.rs/) can generate YAML for GitHub Actions. Let's see how this goes…


sampsyo:
It took a little tweaking, but it works!

PR #181, opened 2024-05-16
Start docs for FlatGFA Python bindings

sampsyo:
Adding to the level of difficulty for CI! We'll see how this goes…


sampsyo:
Looks like it works. Docs are now published here:
https://cucapra.github.io/pollen/flatgfa/

PR #188, opened 2024-07-08
Implement Chop

susan-garry:
Chop works! After `cargo build --release`, try something like `fgfa -I ../tests/k.gfa chop -c 3 -l`. `-c 3` specifies that nodes are to be chopping into segments no longer than 3, and `-l` specifies that the output file should compute new `links` (at this time, it's still not clear to me what need we have for links, if any, but it would be easy to make computing links the default behavior or to always compute links). (Side note, `slow_odgi` does not compute links - do we care to change this?)

The basic algorithm for `chop` is as follows:

```
seg_map;     // map from old segments to their new, chopped counterparts
for each segment:
    chop into segments of size c or smaller
    if args.l:
         link the new segments together, from head to tail (i.e., in the forward orientation)
    update seg_map

for each path:
    new_path;
    for each step in path:
        for new_seg in seg_map(step.seg):
              append new_seg to our new_path
    add new_path to new_fgfa

if args.l:
    for link (A -> B) in old_fgfa:
        add a new link from
             (A.forward ? (A.end, forward) : (A.begin, backwards))
                 -> (B.forward ? (B.begin, forward) : (B.end ? backwards))
```

One weird note here: the implementation of `chop` is split between `cmd.rs` and `main.rs`. The brunt of the work is done in `cmd.rs`, but the logic for which aspects of our original graph to preserve is in `main.rs`. It's unclear that a nice fix exists; because our new graph is borrowing elements from a `GFAStore` created by `chop` in `cmd.rs`, ownership of the `GFAStore` must be passed to the `main` function in order for our new `FlatGFA` to be valid. The best fix may be to compute the `FlatGFA` in `chop` and return both the `FlatGFA` and `GFAStore`, but right now we do not.


.github/workflows/build.yml
```diff
@@ -76,6 +76,32 @@ jobs:
...
+          echo "VIRTUAL_ENV=.venv" >> $GITHUB_ENV
+          echo "$PWD/.venv/bin" >> $GITHUB_PATH
+      - name: Install Python tools
+        run: uv pip install -r requirements.txt
+
+      # Install odgi
+      - name: Pull odgi container
+        run: |
+          docker pull quay.io/biocontainers/odgi:0.8.3--py310h6cc9453_0
+          docker tag quay.io/biocontainers/odgi:0.8.3--py310h6cc9453_0 odgi
```
sampsyo, line 98: Maybe we should make these odgi versions match for the two jobs?

.github/workflows/build.yml
```diff
@@ -43,8 +43,8 @@ jobs:
 
       - name: Pull odgi container
         run: |
-          docker pull quay.io/biocontainers/odgi:0.8.3--py310h6cc9453_0
-          docker tag quay.io/biocontainers/odgi:0.8.3--py310h6cc9453_0 odgi
+          docker pull quay.io/biocontainers/odgi:0.8.6--py310hdf79db3_1
+          docker tag quay.io/biocontainers/odgi:0.8.6--py310hdf79db3_1 odgi
```
sampsyo, line 47: (See next comment.)

Makefile
```diff
@@ -9,14 +9,29 @@ endif
...
 
+fetch-og: $(TEST_FILES:%=tests/%.og)
+
 .PHONY: test-slow-odgi
 test-slow-odgi: fetch
 	make -C slow_odgi test
 
 .PHONY: test-flatgfa
-test-flatgfa:
+test-flatgfa: fetch-og
```
sampsyo, line 25: Might I suggest that we don't actually need to do this `.og` conversion? Seems like odgi works just fine reading directly from a GFA file, and eliminating this step would reduce the surface area of things that can break by one.

bench/bench.py
```diff
@@ -72,7 +72,7 @@ def hyperfine(cmds):
             "hyperfine",
             "--export-json",
             tmp.name,
-            "--shell=none",
+            # "--shell=none",
```
sampsyo, line 75: Fine to just delete this line.

bench/config.toml
```diff
@@ -24,9 +24,14 @@ convert = false
...
 [modes.depth]
 cmd.flatgfa = '{fgfa} -i {files[flatgfa]} depth'
 cmd.odgi = '{odgi} depth -i {files[og]} -d'
-cmd.slow_odgi = '{slow_odgi} depth {files[gfa]}'
\ No newline at end of file
+cmd.slow_odgi = '{slow_odgi} depth {files[gfa]}'
+
+[modes.chop]
+cmd.flatgfa = '{fgfa} -i {files[flatgfa]} chop -c 3'
+cmd.odgi = '{odgi} chop -i {files[og]} -c 3 -o - | {odgi} view -g -i - | {slow_odgi} norm --nl'
```
sampsyo, line 36: For benchmarking purposes, I think we should probably omit the `view` and `norm` steps. This makes for a fairer comparison, where we just measure the time taken to do the data structure transformation and not the time to print out the results. (This of course means that we can't simultaneously check correctness, but I suppose that is probably best done separately.)

flatgfa/src/cmds.rs
```diff
@@ -318,3 +320,199 @@ pub fn depth(gfa: &flatgfa::FlatGFA) {
...
+        }
+    }
+
+    Ok(flat)
+
+    // TODO: Once we figure out how to handle links, fix them?
+    // * Is there any logical correspondence between links and edges?
+    // * Should we preserve/generate updated CIGAR string information? Do we care about links if not?
+    // * * * Maybe generating links/CIGAR strings should be left to whoever is analyzing the graph?
+    // * Go back and add/update links between chopped nodes
```
sampsyo, line 517: Maybe these TODOs are resolved now?

flatgfa/src/flatgfa.rs
```diff
@@ -101,6 +101,12 @@ pub struct Path {
     pub overlaps: Span<Span<AlignOp>>,
 }
 
+impl Path {
+    pub fn step_count(&self) -> usize {
+        self.steps.end.index() - self.steps.start.index()
```
sampsyo, line 106: I think `self.steps.len()` does this. (In fact, maybe we don't need `path.step_count()` if `path.steps.len()` is one character shorter? :smiley:)

flatgfa/src/flatgfa.rs
```diff
@@ -277,8 +283,8 @@ impl<'a> FlatGFA<'a> {
 
     /// Look up a CIGAR alignment.
     pub fn get_alignment(&self, overlap: Span<AlignOp>) -> Alignment {
-        Alignment {
-            ops: &self.alignment[overlap],
+        Alignment { 
+            ops: &self.alignment[overlap]
```
sampsyo, line 287: Can you please run `cargo fmt`? It should get rid of formatting "noise" changes like this one.

flatgfa/src/main.rs
```diff
@@ -104,6 +106,23 @@ fn main() -> Result<(), &'static str> {
         Some(Command::Depth(_)) => {
             cmds::depth(&gfa);
         }
+        Some(Command::Chop(sub_args)) => {
+            let store = cmds::chop(&gfa, sub_args)?;
+            let flat = flatgfa::FlatGFA {
```
sampsyo, line 111: Maybe also deserves a comment about how we'd like to change the encapsulation here someday: to summarize, the problem is that we're creating a "hybrid" `FlatGFA` that borrows from two different underlying stores, and we don't have a good (Rust-safe) way to express that.

flatgfa/src/pool.rs
```diff
@@ -22,6 +22,23 @@ impl<T> Hash for Id<T> {
     }
 }
 
+impl<T> Add<u32> for Id<T> {
+    type Output = Self;
+
+    #[inline]
+    fn add(self, rhs: u32) -> Self::Output {
+        Self(self.0 + rhs, PhantomData)
```
sampsyo, line 30: I'd probably write this as:

```
(self.index() + rhs).into()
```

…which has the benefit of reusing our existing code for `From`.

This probably doesn't matter for this PR, but I'm not 100% sure it is a good idea to support `Add` and `Sub` on `Id`s, as opposed to creating an explicit method like `Id::shift(n)`. The point is that `id + 5` or `id - 3` might look less scary than they should: because they are implicit, they might fool programmers into thinking that something innocuous is happening, whereas doing "index math" like this is actually somewhat advanced. So an explicit method like `shift` might be a better match for forcing programmers to tread carefully when they want to do this kind of index math.

flatgfa/src/print.rs
```diff
@@ -23,6 +23,9 @@ impl fmt::Display for flatgfa::AlignOpcode {
 
 impl<'a> fmt::Display for flatgfa::Alignment<'a> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if self.ops.len() == 0 {
```
sampsyo, line 26: `self.ops.is_empty()`

flatgfa/src/print.rs
```diff
@@ -127,13 +130,13 @@ pub fn write_normalized(gfa: &flatgfa::FlatGFA, f: &mut fmt::Formatter<'_>) -> f
...
+        writeln!(f, "{}", Display(gfa, seg))?;
     }
     for path in gfa.paths.all().iter() {
-        write!(f, "{}", Display(gfa, path))?;
+        writeln!(f, "{}", Display(gfa, path))?;
     }
     for link in gfa.links.all().iter() {
-        write!(f, "{}", Display(gfa, link))?;
+        writeln!(f, "{}", Display(gfa, link))?;
     }
```
sampsyo, line 140: Oops; thanks for fixing this.

tests/turnt.toml
```diff
@@ -170,3 +170,21 @@ output.gfa = "-"
...
+[envs.odgi_depth]
+binary = true
+command = "odgi depth -d -i {filename}"
+output.depth = "-"
+
+[envs.flatgfa_depth]
+command = "../flatgfa/target/debug/fgfa -I {filename} depth"
+output.depth = "-"
+
+[envs.chop_oracle_fgfa]
```
sampsyo, line 183: Maybe this should be called `chop_oracle_with_links` or something to describe it somewhat more specifically?

flatgfa/src/cmds.rs
```diff
@@ -318,3 +320,199 @@ pub fn depth(gfa: &flatgfa::FlatGFA) {
...
+) -> Result<flatgfa::HeapGFAStore, &'static str> {
+
+    let mut flat = flatgfa::HeapGFAStore::default();        
+
+    let mut seg_map: Vec<(Id<Segment>, Id<Segment>)> = Vec::new();
+    let mut max_node_id = 1;
+
+    fn empty_span<T>() -> Span<T> {
+        Span::new(Id::new(0), Id::new(0))
+    }
```
sampsyo, line 355: I think you have a utility method for this now (`Span::new_empty()`)?

flatgfa/src/cmds.rs
```diff
@@ -318,3 +320,199 @@ pub fn depth(gfa: &flatgfa::FlatGFA) {
...
+/// Generates a new graph, rather than modifying the old one in place
+pub fn chop<'a>(
+    gfa: &'a flatgfa::FlatGFA<'a>,
+    args: Chop,
+) -> Result<flatgfa::HeapGFAStore, &'static str> {
+
+    let mut flat = flatgfa::HeapGFAStore::default();        
+
+    let mut seg_map: Vec<(Id<Segment>, Id<Segment>)> = Vec::new();
+    let mut max_node_id = 1;
```
sampsyo, line 351: These could perhaps use comments to describe what they do and what invariants they maintain?

flatgfa/src/cmds.rs
```diff
@@ -318,3 +320,199 @@ pub fn depth(gfa: &flatgfa::FlatGFA) {
...
+                Orientation::Forward => {
+                    // In this builder, Id.index() == seg.name - 1 for all seg
+                    path_end = flat.add_steps(
+                        (start_idx..end_idx).map(|idx| {
+                            Handle::new(
+                                Id::new(idx),
+                                Orientation::Forward
+                            )
+                        })
+                    ).end;
```
sampsyo, line 436: DEFINITELY not for this PR, but: I wonder if there's some kind of utility method that we can invent to help with this stuff. Impressionistically speaking, what we want here is...

```rs
let segs = seg_map[step.segment()];
flat.add_steps(segs.map(|s| Handle::new(s, Orientation::Forward)));
```

Like, we kind of want a way to do a `map` directly on a chunk of segments, without having to fiddle with the index math here. Maybe we can think of a clever way to make that look nice!

flatgfa/src/cmds.rs
```diff
@@ -318,3 +320,199 @@ pub fn depth(gfa: &flatgfa::FlatGFA) {
...
+                            Orientation::Forward
+                        )
+                    },
+                    Orientation::Backward => {
+                        Handle::new(
+                            chopped_segs.0,
+                            Orientation::Backward
+                        )
+                    }
+                }
```
sampsyo, line 483: Could be a little shorter/clearer as:

```rs
let seg_id = match old_from.orient() {
  Orientation::Forward => chopped_segs.1 - 1,
  Orientation::Backward => chopped_segs.0,
};
Handle::new(seg_id, old_from.orient())
```

flatgfa/src/cmds.rs
```diff
@@ -318,3 +320,199 @@ pub fn depth(gfa: &flatgfa::FlatGFA) {
...
+                            Orientation::Forward
+                        )
+                    },
+                    Orientation::Backward => {
+                        Handle::new(
+                            chopped_segs.1 - 1,
+                            Orientation::Backward
+                        )
+                    }
+                }
```
sampsyo, line 501: A similar simplification is probably possible here.

sampsyo (commented):
AWESOME!! This looks really great. I have just a few low-level suggestions, if you're interested!

flatgfa/src/cmds.rs
susan-garry: Makes sense! `seg_map` currently maps from indexes (or segment names) to a tuple of `Id`s, which preserves some type safety but could be better (if I wrap `Segment.name` in an `Id`, it would provide more type safety but I think I'd have to use a HashMap instead of a vector. Maybe there's a trait that could be implemented to automatically convert `Id`s to `ints`? Also, I had been thinking of the `Id` wrapper as a zero cost abstraction, but actually, wouldn't it need to be stored on the heap instead of the stack, and wouldn't this add overhead in some cases?).

Anyways, this probably means that doing a `map` directly on a chunk of segments is not particularly helpful here, since it would still entail mapping from indexes to segments. However, I think that implementing `Range` or something similar, such that we can write something like `start_id..end_id.map(|id| { Handle::new(Id, Orientation::Forward })` would help make this function more readable. In general, it seems like being able to treat `Id`s like regular numbers is a good thing (as long as they're distinct to the type checker). Something like this could probably be implemented in a small follow-up PR?

.github/workflows/build.yml
susan-garry: Done! Changed this to the newer version

Makefile
susan-garry: Aah, makes sense! I left the `fetch-og` target because it is helpful when benchmarking

bench/config.toml
susan-garry: Yep, you're absolutely right. This explains the weird benchmarking results that I was getting for `chop`. Thanks for catching it!

flatgfa/src/cmds.rs
susan-garry: I have replaced this and the other calls to `empty_span` with the utility method

flatgfa/src/cmds.rs
susan-garry: Aah, yep - I have added comments. `max_node_id` is actually the maximum `node_id` currently in existence, +1, so maybe I should pick a better name for it...

flatgfa/src/cmds.rs
sampsyo: Maybe this is silly, but could the RHS of `seg_map` contain `Span<Segment>` instead of `(Id<Segment>, Id<Segment>)`? Then we could perhaps implement a `map` function on `Span`s...

PR #203, opened 2025-03-11
Trivial GAF List Bindings

mt-xing:
Introduces new Python bindings to the `flatgfa-py` project that allows one to trivially turn a GAF parse into a Python list of list of `PyChunkEvent` objects, which can then be iterated over.

Sample code:
```
graph = flatgfa.load("test.flatgfa")
print(graph.size)

for read in graph.test_gaf("test.gaf"):
    for event in read:
        print(event.handle)
        print(event.range)
        print(event.get_seq())
```

Note that `handle`, `range`, and `get_seq()` all return strings for now.


flatgfa-py/src/lib.rs
```diff
@@ -126,6 +127,97 @@ impl PyFlatGFA {
...
+struct PyChunkEvent {
+    chunk_event: Arc<ChunkEvent>,
+    gfa: Arc<Store>,
+}
+
+#[pymethods]
+impl PyChunkEvent {
+    fn __str__(&self) -> String {
+        "".to_string()
+    }
```
sampsyo, line 188: This doesn't seem like it adds much; we can probably just leave it out. :smiley:

flatgfa-py/src/lib.rs
```diff
@@ -126,6 +127,97 @@ impl PyFlatGFA {
         mmap.flush()?;
         Ok(())
     }
+    
+    #[getter]
+    fn size(&self) -> usize {
+        let gfa = self.0.view();
+        file::size(&gfa)
+    }
```
sampsyo, line 135: Maybe I'm missing something, but this seems to be unused?

flatgfa-py/src/lib.rs
```diff
@@ -126,6 +127,97 @@ impl PyFlatGFA {
...
+        for read in parser {
+            print!("{}\t", read.name);
+            for event in flatgfa::ops::gaf::PathChunker::new(&gfa, &name_map, read) {
+                event.print_seq(&gfa);
+            }
+            println!();
+        }
+    }
+
+    fn test_gaf(&self, gaf: &str) -> Vec<Vec<PyChunkEvent>> {
```
sampsyo, line 155: Seems like this could have a more descriptive name: like, `all_reads` or something.

flatgfa-py/src/lib.rs
```diff
@@ -126,6 +127,97 @@ impl PyFlatGFA {
...
+}
+
+#[pymethods]
+impl PyChunkEvent {
+    fn __str__(&self) -> String {
+        "".to_string()
+    }
+
+    #[getter]
+    fn handle(&self) -> String {
```
sampsyo, line 191: This could return an actual `PyHandle` instead of a string representation of the handle.

flatgfa-py/src/lib.rs
```diff
@@ -126,6 +127,97 @@ impl PyFlatGFA {
...
+        let seg_num: u32 = self.chunk_event.handle.segment().into();
+        format!("{}, {}", &self.chunk_event.handle.orient(), seg_num)
+    }
+
+    #[getter]
+    fn range(&self) -> String {
```
sampsyo, line 197: Similarly, this could return a pair of Python integers instead of the stringified version.

flatgfa-py/src/lib.rs
```diff
@@ -126,6 +127,97 @@ impl PyFlatGFA {
...
+        match self.chunk_event.range {
+            flatgfa::ops::gaf::ChunkRange::None => { "".to_string() },
+            flatgfa::ops::gaf::ChunkRange::All => { "all".to_string() },
+            flatgfa::ops::gaf::ChunkRange::Partial(start, end) => {
+                format!("[{}, {})", start, end)
+            }
+        }
+    }
+
+    fn get_seq(&self) -> String {
```
sampsyo, line 207: For symmetry with `PySegment`, this could be called `sequence`.

sampsyo (commented):
Sorry for being so slow at giving this a careful read-through! It overall looks good. We had talked about adding some tests, from the Python side, which would be especially useful for checking that the "fancy" version of this binding works the same way—I think that is the only real thing missing here. I also have a couple of suggestions for naming and things.

@mt-xing, do still you feel like giving the tests a shot? I might be able to try as well…

flatgfa-py/src/lib.rs
mt-xing: I wasn't quite sure which convention to use for the `None` and `All` case. I assume the range is inclusive on both ends, so I made `All` return `(0, size - 1)` while `None` returns `(1, 0)`, but I can change that if we want.

sampsyo:
This looks just about right! There are a couple of minor Clippy warnings, which I'll resolve now and then merge. (Then I can ship the little test I made for this!)

sampsyo:
The last few fixes for Clippy warnings were in f26337e07aa592bf54fdd7b951c693c8d7127903. @mt-xing, it might be handy to take a look and think briefly about why the simplifications are valid, as part of your ongoing Rust-learning quest.

PR #207, opened 2025-04-08
Python Binding for handling GAF files.

JerryJi0730:
For this PR, we have added following code:
1. Allowing python binding to access the internal object like "GAFLine, PathChunker, and PathEvent". The available fields and functions are listed in the flatgfa.pyi
2. Adding a new example file "GAFexample.py" to show how to play around the GAF objects in the python bindings. 
3. We tested the Python binding via printing the same results as "fgfa -I chr6.gfa gaf chr6gaf.gaf".  After testing by hyperfine, we found out that, unsurprisingly, the processing time of Python binding is slower 2 times than the original Rust source code. However, it is probably acceptable if we point out that the processed GAF files and GFA files are both 2 MB large?  
image


flatgfa-py/GAFexample.py
sampsyo, line 1: Let's reorganize our example scripts into a directory. Please create a new directory here called something like `examples`; the old `example.py` can become `depth.py` or similar, and this can become `gaf.py` inside that new directory.

flatgfa-py/chr6.gfa
sampsyo, line 1: Please remove this and the GAF file; we should not commit big data files to git.

flatgfa-py/example.py
```diff
@@ -6,7 +6,6 @@
 for path in graph.paths:
     for step in path:
         depths[step.segment.id] += 1
-
```
sampsyo, line 9: This looks like an accidental change; please revert it.

flatgfa/src/ops/gaf.rs
```diff
@@ -7,7 +7,7 @@ use rayon::iter::{plumbing::UnindexedConsumer, ParallelIterator};
 pub struct GAFLineParser<'a> {
     buf: &'a [u8],
 }
-
+#[derive(Clone)]
 #[derive(Debug)]
```
sampsyo, line 11: This can be written as `#[derive(Debug, Clone)]`.

flatgfa/src/ops/gaf.rs
```diff
@@ -137,18 +137,19 @@ impl<'a, 'b> PathChunker<'a, 'b> {
...
-enum ChunkRange {
+pub enum ChunkRange {
     None,
     All,
     Partial(usize, usize),
 }
 
 impl ChunkEvent {
+
```
sampsyo, line 152: Please run `cargo fmt` to standardize the formatting in your Rust code.

flatgfa/src/ops/gaf.rs
```diff
@@ -191,6 +192,46 @@ impl ChunkEvent {
...
     }
+    pub fn get_seq_string(&self, gfa: &flatgfa::FlatGFA) -> String {
+        let seq = gfa.get_seq_oriented(self.handle);
+    
+        match self.range {
+            ChunkRange::Partial(start, end) => seq.slice(start..end).to_string(),
+            ChunkRange::All => seq.to_string(),
+            ChunkRange::None => String::new(),
+        }
+    }
```
sampsyo, line 203: This duplicates some code with the `print_seq` method. Please deduplicate this so we only have one copy of the `match`… perhaps it would be most straightforward to make `print_seq` call this instead of doing its own `match`? Or maybe we want a third method they both call that returns a `Sequence`?

flatgfa/src/ops/gaf.rs
```diff
@@ -191,6 +192,46 @@ impl ChunkEvent {
...
+    pub fn get_seg(&self, gfa: &flatgfa::FlatGFA)->String {
```
sampsyo, line 204: Similarly, this duplicates code with the existing `print` method. Please combine them into one implementation, with one method calling the other… or maybe we just want the version that returns a string, and the client can just use `println!` on that if it wants to print.

flatgfa-py/src/lib.rs
```diff
@@ -44,6 +51,10 @@ impl Store {
             Store::File(ref mmap) => file::view(mmap),
         }
     }
+    fn py_depth(&self) {
+        let gfa = self.view();
+        depth(&gfa)
+    }
```
sampsyo, line 57: This looks like an unrelated change. Please remove it from this PR so we can focus on the matter at hand: the GAF parser workflow.

flatgfa-py/src/lib.rs
```diff
@@ -85,6 +96,10 @@ impl PyFlatGFA {
         })
     }
 
+    fn depth(&self) {
+        self.0.py_depth()
+    }
```
sampsyo, line 101: As above; let's leave this out.

flatgfa-py/src/lib.rs
```diff
@@ -126,6 +141,17 @@ impl PyFlatGFA {
         mmap.flush()?;
         Ok(())
     }
+    fn load_gaf(&self, gaf: &str) -> PyGAFParser {
+        let gfa = self.0.view();
+        let name_map1 = flatgfa::namemap::NameMap::build(&gfa);
```
sampsyo, line 146: I suggest that `name_map` would be a more logical name than `name_map1`.

flatgfa-py/src/lib.rs
```diff
@@ -444,8 +470,131 @@ impl PyPath {
         self.steps().__len__()
     }
 }
+#[derive(Clone)]
+#[pyclass(frozen)]
+#[pyo3(name = "ChunkEvent", module = "flatgfa")]
+struct PyChunkEvent {
+    chunk_event: Arc<ChunkEvent>,
```
sampsyo, line 477: Because the `ChunkEvent` struct is "just plain data," I think it makes a little more sense to just copy it instead of of using an `Arc` to refer to it. That is, we can add `#derive(Clone)` to `ChunkEvent` and then store a copy here (of type `ChunkEvent`, instead of `Arc<ChunkEvent>`).

flatgfa-py/src/lib.rs
```diff
@@ -444,8 +470,131 @@ impl PyPath {
...
+}
+
+#[pymethods]
+impl PyChunkEvent {
+    #[getter]
+    fn handle(&self) -> String {
```
sampsyo, line 483: Let's make this return a `PyHandle` instead of a string.

flatgfa-py/flatgfa.pyi
```diff
@@ -61,6 +61,26 @@ class LinkList:
...
     def __iter__(self) -> Iterator[Link]: ...
     def __len__(self) -> int: ...
+class ChunkEvent:
+    index:int
+    handle:Handle
+    range: list[int]
+    def __getseq__(self)-> None:...
+    def __getindex__(self)-> None:...
+    def __gethandle__(self)-> None:...
+    def __getrange__(self)-> None:...
```
sampsyo, line 71: It seems like these type declarations are "out of sync" with the actual implementation. For example, there is no actual method called `__getseq__` defined. Can you please make these type declarations align with the implementation in `lib.rs`?

flatgfa-py/src/lib.rs
```diff
@@ -444,8 +470,131 @@ impl PyPath {
...
+        let seg_num: u32 = self.chunk_event.handle.segment().into();
+        format!("{}, {}", &self.chunk_event.handle.orient(), seg_num)
+    }
+
+    #[getter]
+    fn range(&self) -> String {
```
sampsyo, line 489: Similarly, I think it makes a bit more sense for this to return a pair of integers instead of a string.

flatgfa-py/src/lib.rs
```diff
@@ -444,8 +470,131 @@ impl PyPath {
...
+        match self.chunk_event.range {
+            flatgfa::ops::gaf::ChunkRange::None => "".to_string(),
+            flatgfa::ops::gaf::ChunkRange::All => "all".to_string(),
+            flatgfa::ops::gaf::ChunkRange::Partial(start, end) => {
+                format!("[{}, {})", start, end)
+            }
+        }
+    }
+    fn get_seq(&self, gfa: &PyFlatGFA) -> String {
```
sampsyo, line 498: For symmetry with `PySegment`, this should probably be called `sequence`.

flatgfa-py/src/lib.rs
```diff
@@ -444,8 +470,131 @@ impl PyPath {
...
+        let inner_gfa = gfa.0.view();
+        let seq = inner_gfa.get_seq_oriented(self.chunk_event.handle);
+
+        match self.chunk_event.range {
+            flatgfa::ops::gaf::ChunkRange::Partial(start, end) => seq.slice(start..end).to_string(),
+            flatgfa::ops::gaf::ChunkRange::All => seq.to_string(),
+            flatgfa::ops::gaf::ChunkRange::None => "".to_string(),
+        }
```
sampsyo, line 506: I think this logic is already implemented in `ChunkEvent`. Let's just call `get_seq_string` (or whatever it ends up being named) there instead of re-implementing this functionality here.

flatgfa-py/src/lib.rs
```diff
@@ -672,6 +821,8 @@ impl From<ListRef> for LinkList {
 #[pyo3(name = "flatgfa")]
 fn pymod(m: &Bound<'_, PyModule>) -> PyResult<()> {
     m.add_class::<PyFlatGFA>()?;
+    m.add_class::<PyGAFParser>()?;
+    m.add_class::<PyGAFLine>()?;
```
sampsyo, line 825: We're missing the registration for `PyChunkEvent`.

flatgfa-py/src/lib.rs
```diff
@@ -444,8 +470,131 @@ impl PyPath {
...
+                    .collect(),
+                });
+                res
+            }
+            None => None,
+        }
+    }
+}
 
-/// A sequence of :class:`Path` objects.
```
sampsyo, line 448: Please restore this documentation comment.

flatgfa-py/src/lib.rs
```diff
@@ -444,8 +470,131 @@ impl PyPath {
...
+        }
+    }
+}
+
+#[pyclass]
+#[pyo3(name = "GAFLine", module = "flatgfa")]
+
+struct PyGAFLine {
+    store: Arc<Store>,
+    vec_chunk: Vec<PyChunkEvent>,
```
sampsyo, line 515: A somewhat more predictable name for this field would be `chunks`.

flatgfa-py/src/lib.rs
```diff
@@ -444,8 +470,131 @@ impl PyPath {
...
+    gaf: String,
+}
+#[pymethods]
+impl PyGAFLine {
+    #[getter]
+    fn get_name(&self) -> String {
```
sampsyo, line 521: Because this is already marked as a getter, it is accessed on the Python side as `line.get_name` instead of `line.get_name()`. Therefore, a more logical name would just be plain `name`, so that it looks like a normal field access.

flatgfa-py/src/lib.rs
```diff
@@ -444,8 +470,131 @@ impl PyPath {
...
+        self.gaf.clone()
+    }
+
+    #[getter]
+    fn get_chunk_list(&self) -> Vec<PyChunkEvent> {
```
sampsyo, line 526: Similarly, let's call this `chunks`.

flatgfa-py/src/lib.rs
```diff
@@ -444,8 +470,131 @@ impl PyPath {
...
+        self.vec_chunk.clone()
+    }
+    fn get_sequence(&self) -> String {
```
sampsyo, line 529: Let's try to be consistent about what we call the method that gets the underlying sequence data from an object. Elsewhere (e.g., in `PySegment`), it's called `sequence()`, so staying consistent with that would make sense.

flatgfa-py/src/lib.rs
```diff
@@ -444,8 +470,131 @@ impl PyPath {
...
+        let gfa = self.store.view();
+        let mut res: String = "".to_string();
+        for part in self.vec_chunk.clone() {
+            res = res + &part.chunk_event.get_seq_string(&gfa);
+        }
+        res
+    }
+    fn get_seg(&self) -> String {
```
sampsyo, line 537: This also probably needs a more logical name: something like `segment_ranges` or something, to indicate that it (somewhat idiosyncratically) yields a human-readable description of all the various chunks in the line. It is really only useful for debugging, and the name should reflect that.

flatgfa-py/src/lib.rs
```diff
@@ -444,8 +470,131 @@ impl PyPath {
...
+    }
+
+    #[getter]
+    fn get_chunk_list(&self) -> Vec<PyChunkEvent> {
+        self.vec_chunk.clone()
+    }
+    fn get_sequence(&self) -> String {
+        let gfa = self.store.view();
+        let mut res: String = "".to_string();
+        for part in self.vec_chunk.clone() {
```
sampsyo, line 532: The `clone` here seems unnecessary, and it's also expensive: it creates a copy of the entire vector, only to throw it away again immediately. Can we get away without it?

sampsyo (changes requested):
This seems great overall; nice work getting this to go all the way through!! I have a few low-level suggestions throughout the PR—please let me know if anything isn't clear and I can explain more. :smiley:

flatgfa-py/flatgfa.pyi
JerryJi0730: You are right, the declarations are not accurate ai this point. But I believe the only method "exists" here is the "__getseq__", the others are not actual methods.

sampsyo (changes requested):
Looking good overall! Could you please do a couple of small tasks to get ready for this to merge?

1. Run `cargo fmt` and `cargo clippy`.
2. Match the current behavior of the recently-merged #203, including returning a pair of integers from the `range` function.
3. Check that my new tests from #211 still pass with this new approach. (I think exactly the same API should work?) (I have one annoying Python/Windows error to resolve in that PR, but then it should be good to go.)

JerryJi0730:
> 3. Check that my new tests from [Add tests for Python GAF bindings #211](https://github.com/cucapra/pollen/pull/211) still pass with this new approach. (I think exactly the same API should work?) (I have one annoying Python/Windows error to resolve in that PR, but then it should be good to go.)

Thank you for the comments! I am not sure what does the "same API" refer to. Do you refer the "all_reads" function? In my implementation, it is quite different to get the sequences, so it probably looks like two separate api calls. For example, To get  the sequences in a single ChunkEvent, I should iterate over all GafLines from gfa.load_gaf, and then iterate the GAFLines to get each chunkEvent. Should I add one more test but with different ways to get the sequences and ranges?

sampsyo:
> For example, To get the sequences in a single ChunkEvent, I should iterate over all GafLines from gfa.load_gaf, and then iterate the GAFLines to get each chunkEvent.

Hmm… TBH, I actually thought this worked the same way in both implementations? For example, consider this line from my new test (which works with the "old" bindings):

```py
seqs = ["".join(e.sequence() for e in line) for line in gaf]
```

That works by:

* Iterating over each line in the GAF file.
* Iterating over each `ChunkEvent` in that line.
* Calling `sequence` on the `ChunkEvent`.

…which I think is how your "new" bindings would work as well? The only difference is whether `for line in gaf` iterates over a list vs. an iterator, and similarly for `for e in line`. In the current version of your PR, you'd have to use `line.chunks` instead of just `line`, but it's pretty close—in fact, you could add an `__iter__` method to make it look exactly the same. I hope I'm reading this correctly!

sampsyo:
Great; thanks! Yes, let's resolve those conflicts next.

Your example is a good one:

> For example, in my version of all_reads, I will return a GAFLine instead of chunkevents.

I guess the point I want to make here is that, from the perspective of Python iteration, _these two return types are the same_. That is, the old return type was `list[list[ChunkEvent]]`. The new return type is `GAFParser`. But the latter type is an iterable sequence of iterable sequences of `ChunkEvent`s—so it amounts to the same thing, as long as all the client does is iterate over the containers.

So I think we have two options for what to do with the `.pyi` file:

1. Just use the new type, i.e., `GAFParser`. But the tests should not need to change at all.
2. Use [`collections.abc.Iterable`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable), as in `Iterable[Iterable[ChunkEvent]]`. This would be an abstract type that encompasses both.

In both cases, this is *just* about the `.pyi` interface types; the actual implementation (in Rust) need not change. Does this make sense?

@mt-xing, now would be a great time to take a closer look at @JerryJi0730's code here and weigh in!

sampsyo:
Very nice; looks great from here!! I'm going to hit the merge button. But here is another ping for @mt-xing; it would be really great to take a look and offer any comments you have on @JerryJi0730's approach here!

PR #208, opened 2025-04-10
Sub-byte sequence compression 

johnpalsberg:
For this pull request, I added a new file, packseq.rs, in flatgfa/src which contains data structures for storing sub-byte compressed nucleotide sequences in a vector-like format. 

The current two data structures are PackedVec and PackedSeq. An instance of PackedVec stores the underlying sequence in a compressed format (2 base pairs per byte), while an instance of PackedSeq is a reference to a  subsection of a PackedVec.


flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,293 @@
...
+}
+
+impl PackedVec {
+    fn new() -> Self {
+        PackedVec {
+            data: Vec::new(),
+            high_nibble_end: false,
+        }
+    }
+    pub fn push(&mut self, value: u8) {
```
sampsyo, line 26: Let's add an assertion here that `value` is within a safe range that it can be stored in a nibble. Maybe something like `assert!(value <= 0xF)`?

flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,293 @@
...
+// impl Index<usize> for PackedVec {
+//     type Output = u8;
+//     fn index(&self, index: usize) -> &Self::Output {
+//         if index % 2 == 1 {
+//             (&self.data[index as usize / 2] & 0b11110000u8) >> 4
+//         } else {
+//             &self.data[index as usize / 2] & 0b00001111u8
+//         }
+//     }
+// }
```
sampsyo, line 48: It's OK to delete this unused code.

flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,293 @@
...
+            data: &self.data,
+            cur_index: 0,
+            cur_high_nibble: false,
+            high_nibble_end: self.high_nibble_end,
+            last: false
+        };
+        let mut i = 0;
+        for item in iter {
+            if i == 0 {
+                write!(f, "{}", match item {0 => 'A', 1 => 'C', 2 => 'G', 3 => 'T', _ => ' '})?;
```
sampsyo, line 63: This kind of conversion to characters could be simplified by using a `Nucleotide` enum, as outlined above.

flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,293 @@
...
+
+
+/// Used by PackedSlices to reference a subsection of a nucleotide sequence  
+struct Span {
+    start: u8,
+    end: u8,
+}
+
+
+fn get_seq(compressed_data: &Vec<u8> , span: Span)  -> Vec<u8> {
```
sampsyo, line 122: For functions like this, I think it would make sense to move them to methods in the `impl PackedVec` implementation. This would make it easier to share functionality and would keep the utilities clearly organized around the data structure they work with. Do you think that seems feasible?

flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,293 @@
...
+fn print_arr(arr: Vec<u8>) {
+    let new_arr: Vec<char> = arr
+        .into_iter()
+        .map(|e| match e {
+            0 => 'A',
+            1 => 'C',
+            2 => 'T',
+            3 => 'G',
+            _ => ' ',
+        })
```
sampsyo, line 217: This is another place where centralizing some logic around an `enum` for the base-pair symbols would be helpful and could reduce some duplication.

flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,293 @@
...
+
+fn print_arr(arr: Vec<u8>) {
+    let new_arr: Vec<char> = arr
+        .into_iter()
+        .map(|e| match e {
+            0 => 'A',
+            1 => 'C',
+            2 => 'T',
+            3 => 'G',
+            _ => ' ',
```
sampsyo, line 216: I say let's make this an error: as in, use `panic!` to crash if we ever get an unsupported value.

sampsyo (changes requested):
Awesome; this all seems like a great start! My main organizational suggestions are:

* Use an `enum` for the A/C/T/G nucleotide symbols to centralize some logic around converting to/from numbers and readable characters.
* Take the top-level functions that currently deal with raw `u8`s and make them into methods on the `PackedVec` struct. This will help keep things organized and make it easier to use the API correctly.

flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,328 @@
...
+            let c: char = n.into();
+            println!("CHAR!!! {}", c);
+            write!(f, "{}", c)?;
+        }
+        write!(f, "]")
+    }
+}
+
+struct PackedVecIterator<'a> {
+    data: &'a Vec<u8>,
```
sampsyo, line 100: I want to suggest a big simplification of the way iterators work. Instead of taking a reference to the underlying `Vec<u8>`, instead take a reference to the entire `PackedVec`. Then, add the following utilities to the `PackedVec` impl:

* `len()`, the number of nucleotides in the vector.
* `get(index: usize)`, which is basically `get_vec_elem`.

Then, this iterator's logic can be substantially simplified. You don't need to keep track of nibble-related issues; you just need to track the current index and stop when you reach `self.vec.len()`. Does this make sense?

flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,328 @@
...
+            return None;
+        }
+        if self.cur_index == self.data.len() - 1 && !self.high_nibble_end {
+            self.last = true;
+        }
+        let result = if self.cur_high_nibble {
+            (self.data[self.cur_index] >> 4) & 0b00001111u8
+        } else {
+            self.data[self.cur_index] & 0b00001111u8
+        };
```
sampsyo, line 133: Per the above comment, this duplicates the logic in `get_vec_elem`. Using a simpler iterator approach, you can reuse a single implementation of this bit-slicing code.

flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,328 @@
...
+}
+
+/// Returns a compressed PackedVec given an uncompressed vector *arr*
+fn create_vec(arr: Vec<Nucleotide>) -> PackedVec {
+    let mut new_vec = PackedVec::new();
+    for i in 0..arr.len() {
+        new_vec.push(arr[i]);
+    }
+    return new_vec;
+}
```
sampsyo, line 203: I think all these functions would be most naturally expressed as methods in the `PackedVec` impl. Please give that a try—it should admit shorter names for all the functions too (like just plain `get` and `set` instead of `get_vec_elem` and `set_vec_elem`).

flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,328 @@
...
+    for i in span.start..=span.end {
+        arr.push(get_vec_elem(vec_ref, i));
+    }
+    return arr;
+}
+
+/// Returns the element of *seq* at index *index*
+fn get_vec_elem(seq: &PackedVec, index: u8) -> Nucleotide {
+    if index % 2 == 1 {
+        let i: usize = (index / 2) as usize;
```
sampsyo, line 161: You could move this outside of the `if` to avoid duplicating it.

flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,328 @@
...
+        return (&seq.data[i] & 0b00001111u8).into();
+    }
+}
+
+/// Sets the element of *seq* at index *index* to *elem*
+fn set_vec_elem(seq: &mut PackedVec, index: u8, input: Nucleotide) {
+    let elem: u8 = input.into();
+    if index % 2 == 1 {
+        let i: usize = (index / 2) as usize;
```
sampsyo, line 173: As above, it would work to set this outside of the `if` to avoid needing two copies.

flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,328 @@
...
+        println!("i: {}", i);
+        seq.data[i] = (0b00001111u8 & seq.data[i]) | (elem << 4);
+    } else {
+        let i: usize = (index / 2) as usize;
+        seq.data[i] = (0b11110000u8 & seq.data[i]) | elem;
+    }
+}
+
+/// Returns a uncompressed vector that contains the sequence in *seq*
```
sampsyo, line 182: This is literally the least important thing in the world, but because these comments are usually written with Markdown, you probably want to use backticks instead of asterisks to mark references to variable names.
```suggestion
/// Returns a uncompressed vector that contains the sequence in `seq`
```

flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,328 @@
...
+fn get_vec_seq(seq: &PackedVec) -> Vec<Nucleotide> {
+    let end_index = if (seq.high_nibble_end) {
+        (seq.data.len() * 2) - 1
+    } else {
+        (seq.data.len() * 2) - 2
+    };
+    let span: std::ops::Range<u8> = std::ops::Range {
+        start: 0,
+        end: end_index as u8,
```
sampsyo, line 191: If you add a `len` method to `PackedVec`, this could just be `end: self.len()`.

flatgfa/src/packedseq.rs
```diff
@@ -0,0 +1,328 @@
...
+    };
```
sampsyo, line 192: Instead of using an explicit struct constructor here, Rust has special syntax for making `Range` objects! The expression `x..y` is equivalent to `Range { start: x, end: y }`. So you can probably just make this `0..self.len()`.

sampsyo (changes requested):
Very cool so far!! I have a few more Rusty coding suggestions in here—please let me know if anything isn't clear.

sampsyo:
Hi, @johnpalsberg! In the interest of expediency, I took care of the remaining Clippy lints. But please go look at the last few commits on this thread for a few tips on Rust programming in general! One thing that came up a few times, for instance, is that the `return` keyword is rarely necessary; the "last expression in the function" is implicitly its return value.

Issue #206, opened 2025-03-20
Bring flattening to BED files


sampsyo:
Let's take the same "data structure flattening" approach as FlatGFA and apply it to [BED files](https://bedtools.readthedocs.io/en/latest/content/general-usage.html). This would serve a few goals:

* Possibly an easy win that biologists would care about.
* Further prove our general principles by applying it to a second data format.
* Help develop general principles and library functionality for developing flat versions of _arbitrary_ file formats!
* Dovetail nicely with FlatGFA itself, because some of the input/output datasets involved in odgi-style analyses are in fact BED files.
* Get closer to demonstrating the advantages of avoiding serialization/deserialization to text by composing odgi-style and bed tools-style operators.

As a high-level vision, let's take this shell stanza from [this odgi tutorial](https://odgi.readthedocs.io/en/latest/rst/tutorials/detect_complex_regions.html):

```sh
odgi depth -i chr8.pan.og -r chm13#chr8 | \
    bedtools makewindows -b /dev/stdin -w 5000 > chm13.chr8.w5kbps.bed
odgi depth -i chr8.pan.og -b chm13.chr8.w5kbps.bed --threads 2 | \
    bedtools sort > chr8.pan.depth.w5kbps.bed
```

…and port it to FlatGFA-augmented Python:

```py
chr8 = flatgfa.load("chr8.flatgfa")
windows = chr8.depth(path="chm13#chr8").makewindows(5000)
res = chr8.depth(ranges=windows).sort()
res.top(5).print()
```

The type of `chr8` in this fantasy example is `FlatGFA`. The type of `windows` and `res` is something like `FlatBED`. We will need a binary format for this, some interval-oriented operations on this format, and some Python bindings.

Here is a step-by-step plan, broken down into phases.

## 1: Explore

To understand what’s going on here:

- Follow [that odgi tutorial](https://odgi.readthedocs.io/en/latest/rst/tutorials/detect_complex_regions.html), producing a few BED files.
- Using [the BED file format spec](https://bedtools.readthedocs.io/en/latest/content/general-usage.html) and the examples produced above, think carefully about how much of the BED format we need to support to get started with, how much flexibility we want to build in, etc.

## 2: Parser and Printer

Let’s start by *just* writing a BED parser/printer. Let’s not worry about the in-memory data structure too much at first; we’ll just use something flat-ish (i.e., avoid pointers) and worry about the details later.

[...]

PR #210, opened 2025-04-22
Basic BED File Parser and Intersect

mt-xing:
Add an extremely basic BED parser that allows the user to intersect two 3-column BED files.

Format: `fgfa.exe bed -a file1.bed -b file2.bed`

Output should match the result of `bedtools intersect -a file1.bed -b file2.bed

Behavior is undefined on BED files with more than 3 columns for now.

### DIFF
--- a/flatgfa/src/cli/cmds.rs
+++ b/flatgfa/src/cli/cmds.rs
@@ -304,7 +304,6 @@ pub fn gaf_lookup(gfa: &flatgfa::FlatGFA, args: GAFLookup) {
     }
 }
 
-
 /// parse a BED file
 #[derive(FromArgs, PartialEq, Debug)]
 #[argh(subcommand, name = "bed")]
@@ -312,33 +311,31 @@ pub struct BEDIntersect {
     /// first BED file to intersect
     #[argh(option, short = 'a')]
     first_bed_file_path: String,
-    
+
     /// second BED file to intersect
     #[argh(option, short = 'b')]
     second_bed_file_path: String,
 }
 
 pub fn bed_intersect(args: BEDIntersect) {
-
     let bed_file_path = args.first_bed_file_path;
     let bed_file_path2 = args.second_bed_file_path;
 
-        let file = memfile::map_file(&bed_file_path);
-        let bed_store = BEDParser::for_heap().parse_mem(file.as_ref());
-        let bed = bed_store.as_ref();
-        
-            let file2 = memfile::map_file(&bed_file_path2);
-            let bed_store2 = BEDParser::for_heap().parse_mem(file2.as_ref());
-            let bed2 = bed_store2.as_ref();
+    let file = memfile::map_file(&bed_file_path);
+    let bed_store = BEDParser::for_heap().parse_mem(file.as_ref());
+    let bed = bed_store.as_ref();
 
-            for outer_entries in bed.entries.items() {
-                let intersects = bed2.get_intersects(&bed, outer_entries.1);
-                for val in intersects.iter() {
-                    let name: &BStr = bed2.get_name_of_entry(val);
-                    let start = val.start;
-                    let end = val.end;
-                    println!("{}\t{}\t{}", name, start, end);
-                }
-            }
+    let file2 = memfile::map_file(&bed_file_path2);
+    let bed_store2 = BEDParser::for_heap().parse_mem(file2.as_ref());
+    let bed2 = bed_store2.as_ref();
 
+    for outer_entries in bed.entries.items() {
+        let intersects = bed2.get_intersects(&bed, outer_entries.1);
+        for val in intersects.iter() {
+            let name: &BStr = bed2.get_name_of_entry(val);
+            let start = val.start;
+            let end = val.end;
+            println!("{}\t{}\t{}", name, start, end);
+        }
+    }
 }
--- a/flatgfa/src/cli/cmds.rs
+++ b/flatgfa/src/cli/cmds.rs
@@ -317,6 +317,7 @@ pub struct BEDIntersect {
     second_bed_file_path: String,
 }
 
+/// find intersecting intervals between two BED files
 pub fn bed_intersect(args: BEDIntersect) {
     let bed_file_path = args.first_bed_file_path;
     let bed_file_path2 = args.second_bed_file_path;
--- a/flatgfa/src/cli/cmds.rs
+++ b/flatgfa/src/cli/cmds.rs
@@ -1,9 +1,11 @@
+use crate::flatbed::BEDParser;
 use crate::flatgfa::{self, Segment};
-use crate::memfile::map_file;
+use crate::memfile::{self, map_file};
 use crate::namemap::NameMap;
 use crate::ops;
 use crate::pool::Id;
 use argh::FromArgs;
+use bstr::BStr;
 use rayon::iter::ParallelIterator;
 use std::collections::HashMap;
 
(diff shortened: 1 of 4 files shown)


flatgfa/src/flatbed.rs
```diff
@@ -0,0 +1,191 @@
+use zerocopy::{AsBytes, FromBytes, FromZeroes};
+use crate::memfile::MemchrSplit;
+use std::io::BufRead;
+use atoi::FromRadix10;
+use bstr::BStr;
+use crate::pool::{Pool, Span, Store, HeapStore, FixedStore, Id};
+
+#[derive(Debug, FromZeroes, FromBytes, AsBytes, Clone, Copy)]
```
sampsyo, line 8: Suggested documentation comment: "A single interval from a BED file."

flatgfa/src/flatbed.rs
```diff
@@ -0,0 +1,191 @@
...
+#[repr(packed)]
+pub struct BEDEntry {
+    pub name: Span<u8>,
+    pub start: u64,
+    pub end: u64,
+}
+
+pub struct FlatBED<'a> {
```
sampsyo, line 16: A comment that could work here: "A flat representation of an entire BED file, i.e., a list of named intervals."

flatgfa/src/flatbed.rs
```diff
@@ -0,0 +1,191 @@
...
+    pub name_data: Pool<'a, u8>,
+    pub entries: Pool<'a, BEDEntry>,
+}
+
+impl<'a> FlatBED<'a> {
+    /// Get the base-pair sequence for a segment.
```
sampsyo, line 22: This comment appears to be out of date.

flatgfa/src/flatbed.rs
```diff
@@ -0,0 +1,191 @@
...
+    pub fn get_num_entries(&self) -> usize {
+        self.entries.len()
+    }
+
+    pub fn get_name_of_entry(&self, entry: &BEDEntry) -> &BStr {
+        self.name_data[entry.name].as_ref()
+    }
+
+    pub fn get_intersects(&self, bed: &FlatBED, entry: &BEDEntry) -> Vec<BEDEntry> {
+        self.entries.items()
```
sampsyo, line 32: Using `.items()` iterates over `(key, value)` pairs. I think we're only using the values here, so try iterating over `.all()` instead for slightly simpler code.

flatgfa/src/flatbed.rs
```diff
@@ -0,0 +1,191 @@
...
+            .map(|x| BEDEntry {
+                name: x.1.name,
+                start: if x.1.start < entry.start { entry.start } else { x.1.start },
+                end: if entry.end < x.1.end { entry.end } else { x.1.end },
```
sampsyo, line 36: I'm a little confused about the purpose of this `map` step before we do the `filter`. It seems to me that, taken together, we are expressing a predicate on the pair of intervals. As in, we could write a function `intersect(entry1, entry2) -> bool` that just decides whether two intervals overlap—then, we wouldn't need a `map` or to construct a new `BEDEntry` at all; we could simply use a `filter` that calls `intersect`. Would that work?

flatgfa/src/flatbed.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// Get the base-pair sequence for a segment.
+    pub fn get_num_entries(&self) -> usize {
+        self.entries.len()
+    }
+
+    pub fn get_name_of_entry(&self, entry: &BEDEntry) -> &BStr {
+        self.name_data[entry.name].as_ref()
+    }
+
+    pub fn get_intersects(&self, bed: &FlatBED, entry: &BEDEntry) -> Vec<BEDEntry> {
```
sampsyo, line 31: It seems useful for the doc comment for this function to mention the somewhat weird thing about there being two `FlatBED` objects involved here: `self` and `bed`. The tricky thing is that `entry` is from `bed`, not `self`.

flatgfa/src/flatbed.rs
```diff
@@ -0,0 +1,191 @@
...
+        self.entries.items()
+            .map(|x| BEDEntry {
+                name: x.1.name,
+                start: if x.1.start < entry.start { entry.start } else { x.1.start },
+                end: if entry.end < x.1.end { entry.end } else { x.1.end },
+            })
+            .filter(|x| bed.get_name_of_entry(entry).eq(self.get_name_of_entry(x)) && x.end > x.start)
```
sampsyo, line 38: Interesting that there is a string comparison required here. That seems like it is likely expensive. It's not something to do in this PR, but we could consider preprocessing the names so that the comparison here can just use fast integer equality…

flatgfa/src/flatbed.rs
```diff
@@ -0,0 +1,191 @@
...
+
+impl<'a, P: StoreFamily<'a>> BEDParser<'a, P> {
+    pub fn new(builder: BEDStore<'a, P>) -> Self {
+        Self {
+            flat: builder,
+        }
+    }
+
+    /// Parse a GFA text file from an I/O stream.
+    pub fn parse_stream<R: BufRead>(mut self, stream: R) -> BEDStore<'a, P> {
```
sampsyo, line 146: Maybe we can delete this? Not sure we really need to parse streams… I know I did this for GFA files, but it's really just kinda for show and probably not important because `parse_mem` is likely to be so much faster.

flatgfa/src/flatbed.rs
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        self.flat
+    }
+
+    /// Parse a GFA text file from an in-memory buffer.
+    pub fn parse_mem(mut self, buf: &[u8]) -> BEDStore<'a, P> {
+        for line in MemchrSplit::new(b'\n', buf) {
+            let first_tab_index = line.iter().position(|&x| x == b'\t').unwrap();
+            let name_slice = &line[0..first_tab_index];
```
sampsyo, line 167: FWIW, we have a faster alternative to `str.iter().position(|c| c == const)`: this is what [the `memchr` crate](https://docs.rs/memchr/latest/memchr/) does.

Here's how I did it for the GFA parser, FWIW:
https://github.com/cucapra/pollen/blob/711624f3d3de894289e5ee23df6c564ed63328dc/flatgfa/src/gfaline.rs#L128-L142

Basically, it boils down to having a helper that jumps to the nearest `'\t'`, if any, using `memchr::memchr` and then returns the relevant slice. Having this helper (or just reusing the one from `gfaline.rs`?) can make these parsers a bit more readable because this pattern comes up so often.

flatgfa/src/cli/cmds.rs
```diff
@@ -301,3 +303,42 @@ pub fn gaf_lookup(gfa: &flatgfa::FlatGFA, args: GAFLookup) {
...
+}
+
+pub fn bed_intersect(args: BEDIntersect) {
+
+    let bed_file_path = args.first_bed_file_path;
+    let bed_file_path2 = args.second_bed_file_path;
+
+        let file = memfile::map_file(&bed_file_path);
+        let bed_store = BEDParser::for_heap().parse_mem(file.as_ref());
+        let bed = bed_store.as_ref();
```
sampsyo, line 328: Maybe this stuff could be deduplicated with a helper function?

flatgfa/src/cli/main.rs
```diff
@@ -56,6 +57,13 @@ fn main() -> Result<(), &'static str> {
         }
     }
 
+    // Another special case for parsing BED files,
+    // since we do not parse a GFA file for that.
```
sampsyo, line 61: Yeah, I guess the CLI here is due for a redesign… 😳

sampsyo (changes requested):
Very nice!! It's looking great already, and it's awesome that the data structures are already zerocopy. (Actually using that to read & write files will be the next step.)

Would you mind doing a couple of simple quality-of-life things to this PR?

* Adding a few doc comments, where they matter.
* Run `cargo fmt` and `cargo clippy` (cf #209).

Aside from the stuff we already have planned out in #206, here are some suggestions for things to do in the future (not sure how to order these):

* Try to avoid `collect`ing a vector of intervals for output and instead stream them directly to stdout.
* Avoid the string comparison in the inner loop (I left an inner comment about this too) by somehow pre-computing a deduplicated set of names.
* Because our current algorithm here is an $O(N*M)$ approach (and it probably makes sense to stick with this for a while), maybe we should set up some performance tests that don't reveal that. In particular: set up a little benchmark where the "B" file just contains *one* interval, so $M=1$. Then we're just doing an $O(N)$ iteration on the "A" file. This is obviously an unfair comparison against bedtools but maybe it is one we can win.

flatgfa/src/flatbed.rs
mt-xing: If two intervals only partly intersect, `bedtools` only reports the portion that intersects.
eg: if one interval is `1 - 4000` and the other is `2000 - 6000`, `bedtools` will report `2000 - 4000` as the intersection.
So I'm creating the intervals that only contain intersects with the map. Admittedly it's a bit wasteful to do it before the filter, but it made the filter call a lot simpler this way.

sampsyo: Ah, I see!! Thanks for explaining. Maybe a comment to this effect would help make it clearer what's going on.

sampsyo (approved):
Excellent! Looks great. Looking forward to those next steps on optimization…

PR #212, opened 2025-05-07
Crumb compression

johnpalsberg:
Added a new file, packedseqcrumb, which packs 4 Nucleotide base pairs per byte. The user-facing interface of packedseqcrumb is identical to the original packedseq (now called packedseqnibble).


sampsyo:
Very nice; I'm glad this wasn't too different from the 4-bit version!

Your work here has made me think a bit more at a strategic level. I think it would be worthwhile to try to ask whether we can somehow avoid duplicating code to obtain these two design points. In other words, it seems like there's a really important question to ask here: is it possible to maintain a single implementation of `PackedSeq` stuff and have this work for 4-bit, 2-bit, and arbitrary $N$-bit packing? This could be hard, but if we can do it, there's a big payoff: we can add a feature or fix a bug _once_ and have it apply to any "density" we choose.

Does this goal make sense to you? If so, my proposal here is that we skip the 2-bit case and try to see if we can get an arbitrary $N$-bit packed vector. We can talk about potential ways to address this, but the main tools we have available to us here are [const generics](https://practice.course.rs/generics-traits/const-generics.html) and [macros](https://doc.rust-lang.org/book/ch20-05-macros.html). Let's think about what it would take!

In the long term, I think what we want here is something that lets you use _any_ data-free enum:

```
enum Foo { Bar; Baz; Qux };
let v: PackedVec<Foo> = ...;
```

…and the `PackedVec<Foo>` automatically uses ceil(log(number of variants in the enum)) bits per element. I found [one crate that attempts to do this](https://github.com/Badel2/enum_vec) that we could use for inspiration.

But as a stepping stone toward doing that, we could just use a single enum but parameterize on the number of bits (with the rest being padding)? What do you think—worth a shot?

PR #214, opened 2025-06-06
John zerocopy

johnpalsberg:
So far I've created a system for writing and reading compressed sequence data from files. I've created a Rust unit test for this, as well as a command-line test case in a Bash file (running code in a new main.rs file). The current merge conflicts seem to stem from me making certain functions (and the Size struct) in file.rs public.

### DIFF
--- a/flatgfa/src/main.rs
+++ b/flatgfa/src/main.rs
@@ -1,0 +1,32 @@
+use std::env;
+
+use flatgfa::{
+    memfile::{self, *},
+    packedseq::{self, *},
+};
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    if args.len() > 2 && args[1] == "import" {
+        let mmap = memfile::map_file(&args[2]); // args[2] is the filename
+        let seq = packedseq::view(&mmap);
+        println!("Sequence: {}", seq);
+    } else if args.len() > 2 && args[1] == "export" {
+        let vec = PackedSeqStore::create(vec![
+            Nucleotide::A,
+            Nucleotide::C,
+            Nucleotide::T,
+            Nucleotide::G,
+        ]);
+        let input = vec.as_ref();
+        let filename = &args[2];
+        let num_bytes = total_bytes(4);
+        let mut mem = map_new_file(filename, num_bytes as u64);
+        let mut buf = vec![0u8; num_bytes];
+        let buf_ref = &mut buf;
+        dump(&input, buf_ref);
+        mem[..buf_ref.len()].copy_from_slice(buf_ref);
+    } else {
+        println!("Incorrect commands provided!");
+    }
+}
--- a/flatgfa/src/packedseq.rs
+++ b/flatgfa/src/packedseq.rs
@@ -1,5 +1,6 @@
 use crate::file::*;
 use crate::flatgfa;
+use crate::memfile::map_new_file;
 use crate::pool::*;
 use crate::FixedFamily;
 use crate::HeapFamily;
@@ -8,7 +9,7 @@ use std::fmt;
 
 use zerocopy::*;
 
-const MAGIC_NUMBER: u64 = 0x0000_0000;
+const MAGIC_NUMBER: u8 = 0x12;
 
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
 pub enum Nucleotide {
@@ -75,7 +76,7 @@ pub struct PackedSeqView<'a> {
 #[derive(FromBytes, FromZeroes, AsBytes, Debug)]
 #[repr(packed)]
 pub struct PackedToc {
-    magic: u64,
+    magic: u8,
     data: Size,
     high_nibble_end: Size,
 }
@@ -130,9 +131,9 @@ pub fn dump(seq: &PackedSeqView, buf: &mut [u8]) {
     let rest = write_bump(buf, &toc).unwrap();
     let rest = write_bytes(rest, seq.data).unwrap();
 
-    let slice: &[u8] = if seq.high_nibble_end { &[1] } else { &[0] };
+    let nibble: &[u8] = if seq.high_nibble_end { &[1] } else { &[0] };
 
-    write_bytes(rest, slice).unwrap();
+    write_bytes(rest, nibble).unwrap();
 }
 
 impl<'a> PackedSeqView<'a> {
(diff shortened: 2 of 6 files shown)


flatgfa/src/main.rs
sampsyo, line 1: We actually already have a `main.rs`, under the `src/cli` directory. You'll notice that this is organized around subcommands: for example, `fgfa paths` or `fgfa stats` or `fgfa bed`. Instead of adding a new `main.rs` with a new `main` function, let's just add a new subcommand there! Or perhaps two—we could call them `seq-import` and `seq-export`, for example.

flatgfa/src/packedseq.rs
```diff
@@ -45,60 +56,98 @@ impl From<Nucleotide> for char {
...
+
+    /// True if the final base pair in the sequence is stored at a
+    ///                   high nibble
+    high_nibble_end: bool,
+}
+
+#[derive(FromBytes, FromZeroes, AsBytes, Debug)]
+#[repr(packed)]
+pub struct PackedToc {
+    magic: u8,
```
sampsyo, line 79: I know this is a silly thing, but you probably want the [magic number](https://en.wikipedia.org/wiki/File_format#Magic_number) to be an entire word, i.e., a `u64`. Here are two reasons:
* There are just too many collisions in the space of a `u8`! That is, a `u64` that you pick is likely to be unique; a `u8` that you pick is likely to be the same as someone else's `u8` magic number.
* Using a single byte at the start of the file will make the rest of the data unaligned. If the magic number is a `u64`, then the next field will start on a word-aligned boundary.

flatgfa/src/packedseq.rs
```diff
@@ -45,60 +56,98 @@ impl From<Nucleotide> for char {
...
+        data: data.into(),
+        high_nibble_end: match high_nibble_end {
+            [0u8] => false,
+            [1u8] => true,
+            _ => panic!("Invalid value in high_nibble_end"),
+        },
     }
+}
+
+pub fn dump(seq: &PackedSeqView, buf: &mut [u8]) {
```
sampsyo, line 129: I think this also probably makes sense to be a method on `PackedSeqView`.

flatgfa/src/packedseq.rs
```diff
@@ -204,20 +293,33 @@ pub fn create_slice(vec: &PackedVec, s: std::ops::Range<usize>) -> PackedSlice<'
...
+pub fn total_bytes(num_elems: usize) -> usize {
+    let bytes = std::mem::size_of::<usize>();
+    let seq_size = 1 + if num_elems % 2 == 1 {
+        (num_elems / 2) + 1
+    } else {
+        num_elems / 2
+    }; // Sequence plus the nibble byte
+
+    let toc_size = bytes * 4 + 1; // 2 Size types each with two 8-byte
+                                  // usize types, plus the magic number
```
sampsyo, line 308: You can get the size of the TOC directly using `std::mem::size_of<TOC>`.

flatgfa/src/packedseq.rs
```diff
@@ -204,20 +293,33 @@ pub fn create_slice(vec: &PackedVec, s: std::ops::Range<usize>) -> PackedSlice<'
 
 /// Returns a vector containing the base pairs referenced by `slice`
 pub fn get_slice_seq(slice: PackedSlice<'_>) -> Vec<Nucleotide> {
-    slice.vec_ref.get_range(slice.span)
+    slice.vec_ref.as_ref().get_range(slice.span)
+}
+
+pub fn total_bytes(num_elems: usize) -> usize {
```
sampsyo, line 299: Not 100% sure about this one, but perhaps this API could be a tiny bit friendlier if it took a `PackedSeqStore` or `PackedSeqView` as an argument. It would of course just immediately use `.len()` to get the size, but it would make it a little clearer to callers that they're supposed to use this to get the file size for one of these data structures.

flatgfa/src/pool.rs
```diff
@@ -251,7 +251,7 @@ impl<'a, T> Pool<'a, T> {
         self.0.len()
     }
 
-    /// Check if the pool is empty.
+    /// Check if th e pool is empty.
```
sampsyo, line 254: Looks like a typo!

flatgfa/src/main.rs
```diff
@@ -0,0 +1,32 @@
...
+    memfile::{self, *},
+    packedseq::{self, *},
+};
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    if args.len() > 2 && args[1] == "import" {
+        let mmap = memfile::map_file(&args[2]); // args[2] is the filename
+        let seq = packedseq::view(&mmap);
+        println!("Sequence: {}", seq);
```
sampsyo, line 13: I say let's just print the raw `seq` itself, without the `Sequence: ` prefix. This will make the CLI friendlier for use in larger pipelines.

flatgfa/src/main.rs
```diff
@@ -0,0 +1,32 @@
...
+    } else if args.len() > 2 && args[1] == "export" {
+        let vec = PackedSeqStore::create(vec![
+            Nucleotide::A,
+            Nucleotide::C,
+            Nucleotide::T,
+            Nucleotide::G,
+        ]);
```
sampsyo, line 20: Instead of constructing a specific sequence here, let's read a text file from [standard input](https://doc.rust-lang.org/std/io/struct.Stdin.html)! That is, we'd like to do something like this:

```
$ cat myseq.txt | fgfa seq-export myseq.seq
```

where `myseq.txt` contains text like `ACTGTGGACCAATG` or whatever. In other words, let's make a CLI here that can convert between two file types: text files with one nucleotide per character, and our brand-new binary sequence data format.

flatgfa/src/test_export_import.sh
sampsyo, line 1: A standalone shell script is perfectly fine, and maybe the right tool for the job here. The "next level up" for testing this that I had in mind was to set up [Turnt](https://github.com/cucapra/turnt), along the lines of what we already do for round-trips from GFA to FlatGFA:
https://github.com/cucapra/pollen/blob/8a4fd757078b86e681125f4f1abd9c3c2c1fc340/tests/turnt.toml#L162-L172

Basically, the idea would be to create a little `tests` directory here and to populate it with a few small text sequence files. Then, a Turnt config would look something like this:

```toml
command = "fgfa seq-import < {filename} | fgfa seq-export"
output.txt = "-"
```

This would test that, starting with the original text file, a round trip through importing and exporting (I forget which is which!) eventually produces exactly the same `.txt` file, byte for byte. Would that strategy make sense?

sampsyo (changes requested):
Awesome; this is looking good already! I have a few suggestions around the edges—please let me know where I can clarify further.

sampsyo:
It seems like this PR is now outdated—can we close it?

PR #218, opened 2025-06-19
John-cli

johnpalsberg:
This PR contains a cli for export/import to files, and nothing beyond that (ie. no features beyond the previous suggestions on the john-zerocopy PR)


flatgfa/Cargo.toml
```diff
@@ -17,4 +17,5 @@ memmap = "0.7.0"
 num_enum = "0.7.3"
 rayon = "1.10.0"
 tinyvec = "1.8.0"
+rand = "0.8"
```
sampsyo, line 20: Because `rand` is only used in the tests, and not in the final code itself, can you please make this a dev dependency? https://doc.rust-lang.org/rust-by-example/testing/dev_dependencies.html

flatgfa/src/cli/cmds.rs
```diff
@@ -299,3 +301,50 @@ pub fn gaf_lookup(gfa: &flatgfa::FlatGFA, args: GAFLookup) {
         }
     }
 }
+
+///
```
sampsyo, line 305: Please add a short description of the command. This shows up in the `fgfa help` output, for example. You can see the other commands for a hint about how these are usually formatted (super short phrases).

flatgfa/src/cli/cmds.rs
```diff
@@ -299,3 +301,50 @@ pub fn gaf_lookup(gfa: &flatgfa::FlatGFA, args: GAFLookup) {
...
+
+pub fn seq_import(args: SeqImport) {
+    let vec = packedseq::import(&args.filename);
+    let _ = std::fs::remove_file(&args.filename);
+    let store = PackedSeqStore::create(vec);
+    let view = store.as_ref();
+    print!("{}", view);
+}
+
+///
```
sampsyo, line 321: As above.

flatgfa/src/cli/cmds.rs
```diff
@@ -299,3 +301,50 @@ pub fn gaf_lookup(gfa: &flatgfa::FlatGFA, args: GAFLookup) {
...
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "seq-import")]
+pub struct SeqImport {
+    #[argh(positional)]
+    filename: String,
+}
+
+pub fn seq_import(args: SeqImport) {
+    let vec = packedseq::import(&args.filename);
+    let _ = std::fs::remove_file(&args.filename);
```
sampsyo, line 315: Are you sure you want to delete the file after reading it? It seems like it would be more predictable if we just left it alone, and the user can delete it if they want.

flatgfa/src/cli/cmds.rs
```diff
@@ -299,3 +301,50 @@ pub fn gaf_lookup(gfa: &flatgfa::FlatGFA, args: GAFLookup) {
...
+    let store = PackedSeqStore::create(vec);
+    let view = store.as_ref();
+    print!("{}", view);
+}
```
sampsyo, line 319: It seems like this short function, which is for decompressing a packed sequence file, actually does an extra round trip:

* `import`: Gets the compressed data from disk, and decompresses into an in-memory vector.
* `PackedSeqStore::create`: Compresses uncompressed data into the packed format.
* `PackedSeqView::fmt`: Decompresses the data again so we can format it for the console.

That is, we decompress, recompress, and then decompress again. Would it work to instead just provide a way to get a `PackedSeqView` directly from a file (that's probably what we want, as opposed to what `import` currently does) and then `print!` that?

flatgfa/src/cli/cmds.rs
```diff
@@ -299,3 +301,50 @@ pub fn gaf_lookup(gfa: &flatgfa::FlatGFA, args: GAFLookup) {
...
+    let vec: Vec<packedseq::Nucleotide> = input
+        .chars()
+        .filter(|c| !c.is_whitespace())
+        .map(|c| match c {
+            'A' => packedseq::Nucleotide::A,
+            'C' => packedseq::Nucleotide::C,
+            'T' => packedseq::Nucleotide::T,
+            'G' => packedseq::Nucleotide::G,
+            _ => panic!("Invalid character in fle: {}", c),
+        })
```
sampsyo, line 344: What do you think about providing this functionality in an `impl From<char> for Nucleotide`? Seems like it would fit nicely with all the other various `From` implementations that `Nucleotide` already has.

flatgfa/src/cli/cmds.rs
```diff
@@ -299,3 +301,50 @@ pub fn gaf_lookup(gfa: &flatgfa::FlatGFA, args: GAFLookup) {
...
+        .collect();
+
+    let store = packedseq::PackedSeqStore::create(vec);
```
sampsyo, line 347: I suggest that we probably want a method on `PackedSeqStore` that does this conversion, from a string or `&[u8]` to compressed data. As an added bonus, it could avoid allocating a `Vec<Nucleotide>` at all. Consider something like this:

I suggest that we probably want a method on `PackedSeqStore` that does this conversion, from a string or `&[u8]` to compressed data. As an added bonus, it could avoid allocating a `Vec<Nucleotide>` at all. Consider something very roughly like this:

```rs
fn append_text(&mut self, text_data: &[u8]) {
    for c in text_data {
        self.push(c.into());
    }
}

fn from_text(text_data: &[u8]) -> Self {
    let mut store = Self::new();
    store.append_text(text_data);
    store
}
```

flatgfa/src/cli/main.rs
```diff
@@ -56,6 +58,15 @@ fn main() -> Result<(), &'static str> {
         }
     }
 
+    if let Some(Command::SeqExport(sub_args)) = args.command {
```
sampsyo, line 61: Please add a comment here mentioning the "shortcut" for this case, which inputs or outputs raw sequence data instead of GFA files.

flatgfa/src/packedseq.rs
```diff
@@ -1,5 +1,18 @@
+use crate::file::*;
+use crate::flatgfa;
+use crate::memfile;
+use crate::memfile::map_new_file;
+use crate::pool::*;
+use crate::FixedFamily;
+use crate::HeapFamily;
+use crate::StoreFamily;
```
sampsyo, line 8: Looks like these imports (and perhaps some others) are unused. You can run `cargo build` to see a list of warnings like this.

flatgfa/src/packedseq.rs
```diff
@@ -45,60 +58,88 @@ impl From<Nucleotide> for char {
...
-        let value = input.into();
-        assert!(value <= 0xF);
-        if self.high_nibble_end {
-            self.data.push(value);
-            self.high_nibble_end = false;
-        } else {
-            let last_index = self.data.len() - 1;
-            self.data[last_index] |= value << 4;
-            self.high_nibble_end = true;
+fn read_packed_toc(data: &[u8]) -> (&PackedToc, &[u8]) {
```
sampsyo, line 103: It seems like this should maybe be a method on `PackedToc`, with `Self` in the return type. It could then have a simpler name, like `read`.

flatgfa/src/packedseq.rs
```diff
@@ -45,60 +58,88 @@ impl From<Nucleotide> for char {
...
+    let toc = PackedToc::ref_from_prefix(data).unwrap();
+    let rest = &data[size_of::<PackedToc>()..];
+    let magic = toc.magic;
+    assert_eq!(magic, MAGIC_NUMBER);
+    (toc, rest)
+}
+
+impl<'a> PackedSeqView<'a> {
+    pub fn view(data: &'a [u8]) -> Self {
```
sampsyo, line 112: This could use a somewhat more descriptive name, or a detailed documentation comment, or preferably both. It seems important to give the context that this has to do with storing and reading stuff from files, so `&[u8]` has to be a complete file in our format: i.e., start with a TOC followed by the sequence data.

flatgfa/src/packedseq.rs
```diff
@@ -45,60 +58,88 @@ impl From<Nucleotide> for char {
...
+        let (toc, rest) = read_packed_toc(data);
+
+        let (data, _) = slice_prefix(rest, toc.data);
+        Self {
+            data: data.into(),
+            high_nibble_end: match toc.high_nibble_end {
+                0u8 => false,
+                1u8 => true,
+                _ => panic!("Invalid value in high_nibble_end"),
+            },
```
sampsyo, line 122: Maybe it would be slightly cleaner to move this to a method on `PackedToc`, which accesses the underlying `u8` field and returns a `bool`.

flatgfa/src/packedseq.rs
```diff
@@ -45,60 +58,88 @@ impl From<Nucleotide> for char {
...
         }
     }
 
+    pub fn dump(&self, buf: &mut [u8]) {
```
sampsyo, line 126: As above, it would be great to give this a more descriptive name, a documentation comment, or hopefully both.

flatgfa/src/packedseq.rs
```diff
@@ -45,60 +58,88 @@ impl From<Nucleotide> for char {
...
         } else {
             self.data.len() * 2 - 1
         }
     }
-
     pub fn is_empty(&self) -> bool {
         self.data.is_empty()
     }
 
-    /// Returns the element of this PackedVec at index `index`
```
sampsyo, line 101: Please restore this comment.

flatgfa/src/packedseq.rs
```diff
@@ -204,20 +285,46 @@ pub fn create_slice(vec: &PackedVec, s: std::ops::Range<usize>) -> PackedSlice<'
 
 /// Returns a vector containing the base pairs referenced by `slice`
 pub fn get_slice_seq(slice: PackedSlice<'_>) -> Vec<Nucleotide> {
-    slice.vec_ref.get_range(slice.span)
+    slice.vec_ref.as_ref().get_range(slice.span)
+}
+
+pub fn total_bytes(seq: &PackedSeqView) -> usize {
```
sampsyo, line 291: This seems like it should be a method on `PackedSeqView`. It would also be great to clarify (in docs, and/or the function name) that this is about the _file_ size, i.e., the number of bytes you need to hold the result of `dump`.

flatgfa/src/packedseq.rs
```diff
@@ -204,20 +285,46 @@ pub fn create_slice(vec: &PackedVec, s: std::ops::Range<usize>) -> PackedSlice<'
...
+    } else {
+        num_elems / 2
+    };
+
+    let toc_size = std::mem::size_of::<PackedToc>();
+    toc_size + seq_size
+}
+
+pub fn import(filename: &str) -> Vec<Nucleotide> {
+    let mmap = memfile::map_file(filename); // args[2] is the filename
```
sampsyo, line 304: `args[2]` doesn't seem relevant here.

flatgfa/src/packedseq.rs
```diff
@@ -204,20 +285,46 @@ pub fn create_slice(vec: &PackedVec, s: std::ops::Range<usize>) -> PackedSlice<'
...
+        (num_elems / 2) + 1
+    } else {
+        num_elems / 2
+    };
+
+    let toc_size = std::mem::size_of::<PackedToc>();
+    toc_size + seq_size
+}
+
+pub fn import(filename: &str) -> Vec<Nucleotide> {
```
sampsyo, line 303: I touched on this above in `cmds.rs`, but I think we would want such a utility to return a `PackedSeqView`. The current signature requires a `Vec` allocation and a copy of all the data, which seems unnecessary to require for all clients.

flatgfa/src/packedseq.rs
```diff
@@ -204,20 +285,46 @@ pub fn create_slice(vec: &PackedVec, s: std::ops::Range<usize>) -> PackedSlice<'
...
+    PackedSeqView::view(&mmap).get_elements()
+}
+
+pub fn export(seq: PackedSeqView, filename: &str) {
+    let num_bytes = total_bytes(&seq);
+    let mut mem = map_new_file(filename, num_bytes as u64);
+    let mut buf = vec![0u8; num_bytes];
+    let buf_ref = &mut buf;
+    PackedSeqView::dump(&seq, buf_ref);
+    mem[..buf_ref.len()].copy_from_slice(buf_ref);
```
sampsyo, line 314: This entails two copies where one would suffice:

* It allocates a vector and uses `dump` to put all the data into that vector.
* It then copies all that data into `mem`.

Instead, let's just `dump` the data directly into `mem`. Something roughly like this should suffice:

```rs
seq.dump(mem);
```

flatgfa/src/packedseq.rs
```diff
@@ -312,8 +419,62 @@ fn test_getter_setter() {
...
+    let result: &[u8] = &mem;
+    let output = PackedSeqView::view(result);
+    assert_eq!(input.get(0), output.get(0));
+    assert_eq!(input.get(1), output.get(1));
+    assert_eq!(input.get(2), output.get(2));
+    assert_eq!(input.get(3), output.get(3));
+    std::mem::drop(mem);
+    let _ = std::fs::remove_file(filename);
+}
```
sampsyo, line 452: Having through about this a little more, maybe the sensible thing to do here would be to test round-tripping through a `Vec<u8>` but never actually write that stuff to a file. That's almost as good but excludes the actual memory-mapping step, which is unlikely to be a problem. Basically, the idea would be that this "internal" test would test everything _except_ opening and closing files, basically, but everything else.

This would at least avoid the potential annoyance of creating and deleting temporary files.

Would this make sense? It would at least keep this test code a teensy bit simpler, and leave the actual filename handling to the CLI tests if we add those.

flatgfa/src/packedseq.rs
```diff
@@ -312,8 +419,62 @@ fn test_getter_setter() {
...
+        let mut vec: Vec<Nucleotide> = Vec::new();
+        for _ in 0..len {
+            let rand_num = rng.gen_range(0..=3);
+            match rand_num {
+                0 => vec.push(Nucleotide::A),
+                1 => vec.push(Nucleotide::C),
+                2 => vec.push(Nucleotide::T),
+                3 => vec.push(Nucleotide::G),
+                _ => panic!("Incorrect item appended!"),
+            }
```
sampsyo, line 469: It seems like this is more or less what `Nucleotide`'s `From<u8>` impl already does. Can we just call that instead of making a new `match` here?

flatgfa/src/packedseq.rs
```diff
@@ -312,8 +419,62 @@ fn test_getter_setter() {
...
+        }
+        let old_vec = vec.clone();
```
sampsyo, line 471: `vec` itself is never mutated or destroyed, so do we need to clone it here? Can we just reuse `vec` itself instead of creating `old_vec`?

flatgfa/src/packedseq.rs
sampsyo, line 469: Here is a somewhat fancier way to initialize a `Vec<Nucleotide>`, assuming we can use the `From` impl as discussed above:

```rs
let vec: Vec<Nucleotide> = (0..len).into_iter().map(|n| n.into()).collect();
```

sampsyo (changes requested):
Looking good so far! I have a few suggestions throughout—please let me know if there's anything I can expand on.

flatgfa/tests/test.txt
```diff
@@ -0,0 +1 @@
+[A, C, T, G, G]
```
sampsyo, line 1: This is just a note to myself: after landing this PR, maybe it would make sense to have the CLI output the raw nucleotides instead of a comma-separated list. Then the test could be a true round-trip, using the input file as the output file.

flatgfa/src/cli/cmds.rs
```diff
@@ -334,7 +336,48 @@ pub fn bed_intersect(args: BEDIntersect) {
...
+    let mut input = String::new();
+    std::io::stdin()
+        .read_to_string(&mut input)
+        .expect("Stdin read failure");
+
+    let vec: Vec<packedseq::Nucleotide> = input
+        .chars()
+        .filter(|c| !c.is_whitespace())
+        .map(packedseq::Nucleotide::from)
+        .collect();
```
sampsyo, line 378: Not something to address in this PR, but it would be great to figure out how to avoid this `collect`. Relatedly, we should consider setting up some manner of benchmarking to measure the performance impact of that…

flatgfa/src/cli/main.rs
sampsyo: Still needs a comment.

sampsyo (approved):
Looks like we are finally ready to go! I mentioned a few to-do items to address in follow-ups within…

PR #221, opened 2025-07-18
packedseq: Some code-level niceness and several performance improvements

sampsyo:
This is a follow-up to @johnpalsberg's work in #218 on a command-line interface for compressing and decompressing raw nucleotide sequence data. I found a few minor API things I wanted to fix and a range of performance improvements to test out. I'm sure there's more performance to be wrung out, but these were the low-hanging fruit that I noticed.

Please see the individual commits' messages for some details on what I was thinking in each step. One notable change is that the export/import is now a true byte-for-byte round trip, so our Turnt tests can reuse the input file as the output file.

To measure my performance improvements, I needed a bigger file full of DNA sequence data. I obtained a FASTA file from [the Cosmic project](https://cancer.sanger.ac.uk/cosmic/download/cosmic/v102/genefasta) and stripped out all the FASTA metadata, leaving me with just sequence data. I then further removed all the "N" entries because we don't (yet) handle those. I was left with an 89 MB text file—so not huge, but still enough to measure. I conducted performance measurements using [hyperfine](https://github.com/sharkdp/hyperfine) and my now-geriatric iMac with a 4.2 GHz Core i7 from 2017.

First, here's a log of optimizations on the compression ("export") path:

- initially: exporting took 1.091 s ±  0.015 s
- avoiding the intermediate vector of `Nucleotide`s: 1.016 s ±  0.013 s (roughly 80 ms saved, less than I was hoping)
- avoiding UTF-8 string stuff: 894.7 ms ± 115.1 ms (so that's a bigger savings, about 20%)
- reading from a memory-mapped file instead of stdin: 763.4 ms ±   9.0 ms (probably all from exploiting OS caches better, but I'll still take it)

So in total, the performance win is about 1.4×.

I did much less to the decompression ("import") path:

- initially: 1.369 s ±  0.048 s
- skipping UTF-8 string stuff: 266.0 ms ±  10.0 ms (so roughly 5× faster)

Less effort, but more performance. I guess I'll take it!

### DIFF
--- a/flatgfa/tests/turnt.toml
+++ b/flatgfa/tests/turnt.toml
@@ -1,2 +1,2 @@
-command = "../../target/release/fgfa seq-export {filename} packed.seq ; cargo run seq-import packed.seq ; rm packed.seq"
+command = "../../target/debug/fgfa seq-export {filename} packed.seq ; cargo run seq-import packed.seq ; rm packed.seq"
 output.t = "-"


johnpalsberg (commented):
Looks good. It's cool to see just how much overhead is reduced by going directly to ASCII, and by avoiding stdin. I agree that displaying sequences as raw strings instead of comma-separated lists is good and will simplify things. Thanks for the edits!

sampsyo:
Cool! I'll merge this now.

My best guess is that I was using the release build (i.e., `cargo build --release`) for these tests for a bit so they were fast enough on some big input files. I'll switch this bag to using the debug build. But in the future, try `turnt -vp` to see an actual underlying error message to understand why the tests have failed.

PR #238, opened 2025-11-23
Minimal pangenotype matrix construction

sampsyo:
This PR build on @JerryJi0730's `jerry/pangenotype_martix` branch, which doesn't have a PR yet, and tries to keep things to a minimal size so we can merge it in the short term. Relative to @JerryJi0730's current branch, I have:

* removed dead code
* removed compression dependencies (we can work on this separately)
* moved the core functionality from the Python library into the core FlatGFA library
* added a CLI subcommand to construct the matrix and print it out, named `fgfa matrix`

I hope that this PR can let us focus on carefully reviewing the current version of the code so we can merge this *basic* version of the functionality in the short term. Then, future PRs can add more fanciness (such as performance and dealing with compressed text files), making them easier to review.


flatgfa-py/examples/tiny.gaf
sampsyo: This and `tiny.gfa` seem to duplicate what's in `tests/`, so perhaps we can de-duplicate this.

PR #239, opened 2025-11-24
Minimal Steps by Segment Index PR

ashwinsawant17:
I created the initialization of the index, and some basic get functions for the slice of StepRefs and its length (given by the len() function for Spans).

### DIFF
--- a/flatgfa/src/ops/depth.rs
+++ b/flatgfa/src/ops/depth.rs
@@ -23,14 +23,37 @@ pub fn depth(gfa: &flatgfa::FlatGFA, use_index: bool) -> (Vec<usize>, Vec<usize>
         // Build the index
         let step_seg_index = index::StepsBySegIndex::new(gfa);
 
+        // This bit vector keeps track of whether the current *path* has already been seen
+        // by the current working *segment*. A reverse approach to the non-indexed option
+        let mut path_seen: BitVec = BitVec::from_elem(gfa.paths.len(), false);
+
         // iterate over each segment and populate the output vectors
-        for (id, _) in gfa.segs.items() {
+        for (seg_id, _) in gfa.segs.items() {
+            // clear the path_seen vector, since all paths should be unseen
+            path_seen.clear();
+
             // get the actual offset of the id of thesegment
-            let ind = id.index();
+            let ind = seg_id.index();
+
+            // get the span of StepRefs for this segment
+            let span = step_seg_index.get_steps_slice(seg_id);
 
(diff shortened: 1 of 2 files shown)


sampsyo (changes requested):
Hi, @ashwinsawant17! To bring this to the next step, can you please do a couple of minor "homework" tasks?

1. Split this into a separate file, probably `index.rs` or something. It would be great to make sure this stays distinct from the core FlatGFA data structures.
2. Run `cargo fmt`. Doing this is always a good idea when opening/updating a PR so your reviewers don't get distracted by formatting details.

flatgfa/src/flatgfa.rs
```diff
@@ -550,3 +550,95 @@ pub type FixedGFAStore<'a> = GFAStore<'a, FixedFamily>;
...
+}
+
+
+impl StepsBySegIndex {
+
+    pub fn new(fgfa: &FlatGFA)  -> Self{
+
+
+        /// helper to extract the segment index from the stepref
+        fn segment_of_step(fgfa: &FlatGFA, step: &StepRef) -> usize {
```
sampsyo: You can move this function to the top level, because it doesn't seem to reference the context.

sampsyo: Maybe this would be a little clearer/more type-safe if it returned an `Id<Segment>` instead of a plain `u32`?

sampsyo (changes requested):
Oops, I hit "submit" before including granular code-level comments. Here are some suggestions on the work so far!

flatgfa/src/flatgfa.rs
ashwinsawant17: Moved to the top level in index.rs in the latest commit.

flatgfa/src/ops/depth.rs
```diff
@@ -18,15 +19,53 @@ pub fn depth(gfa: &flatgfa::FlatGFA) -> (Vec<usize>, Vec<usize>) {
...
-    for path in gfa.paths.all().iter() {
-        seen.clear(); // All segments are unseen.
-        for step in &gfa.steps[path.steps] {
-            let seg_id = step.segment().index();
-            depths[seg_id] += 1;
-            if !seen[seg_id] {
-                // The first traversal of this path over this segment.
-                uniq_depths[seg_id] += 1;
-                seen.set(seg_id, true);
+    if use_index {
```
sampsyo: Because these two routes have such completely different implementations, I say let's just put them in separate functions. It will make each one easier to read.

PR #241, opened 2025-12-01
Actually use PackedSeq in FlatGFA

johnpalsberg:
Reduces the size of each SeqSpan from 18 bytes to 8 bytes by using u32 indices and logical indices.

### DIFF
--- a/flatgfa/src/flatgfa.rs
+++ b/flatgfa/src/flatgfa.rs
@@ -457,12 +457,12 @@ impl<'a, P: StoreFamily<'a>> GFAStore<'a, P> {
                 _ => panic!("Not a Nucleotide!"),
             };
             if high_nibble_end {
+                high_nibble_end = false;
                 if i == seq.len() - 1 {
                     self.seq_data.add(converted);
                     break;
                 }
                 combined_item = converted;
-                high_nibble_end = false;
             } else {
                 combined_item |= converted << 4;
                 self.seq_data.add(combined_item);


flatgfa/src/flatgfa.rs
```diff
@@ -354,16 +346,17 @@ impl std::fmt::Display for Sequence<'_> {
...
         Ok(())
     }
 }
 
 impl<'a> FlatGFA<'a> {
     /// Get the base-pair sequence for a segment.
-    pub fn get_seq(&self, seg: &Segment) -> &BStr {
-        self.seq_data[seg.seq].as_ref()
+    pub fn get_seq(&self, seg: &Segment) -> PackedSeqView<'_> {
+        // println!("end {}", seg.seq.end());
```
sampsyo: Looks like a stray debug thing we can remove.

flatgfa/src/flatgfa.rs
```diff
@@ -423,6 +416,71 @@ impl<'a> FlatGFA<'a> {
...
+        self.len == 0
+    }
+
+    /// Given `range`, returns the equivalent SeqSpan
+    pub fn from_range(range: Range<usize>) -> Self {
+        Self {
+            start: range.start as u32,
+            len: (range.end - range.start) as u16,
+        }
+    }
```
sampsyo: Maybe it would be a little more Rust-idiomatic to make this into an `impl From<Range>`. Then you could do `range.into()` and stuff to get the conversion to happen automatically.

flatgfa/src/flatgfa.rs
```diff
@@ -423,6 +416,71 @@ impl<'a> FlatGFA<'a> {
...
+
+    /// Returns the range that is equivalent to this SeqSpan
+    pub fn to_range(&self) -> Range<usize> {
+        Range {
+            start: self.start as usize,
+            end: self.end(),
+        }
+    }
```
sampsyo: Similarly, this could be an `impl From<SeqSpan> for Range<usize>`.

flatgfa/src/flatgfa.rs
```diff
@@ -423,6 +416,71 @@ impl<'a> FlatGFA<'a> {
     }
 }
 
+/// A Span of a packed sequence.
+#[derive(Debug, FromBytes, IntoBytes, Clone, Copy, PartialEq, Eq, Hash, Immutable)]
+#[repr(packed)]
+pub struct SeqSpan {
```
sampsyo: Maybe it would make sense to move `SeqSpan` to `packedseq.rs`, just because it is so tightly coupled with the implementation of `PackedSeqView`?

flatgfa/src/flatgfa.rs
```diff
@@ -448,9 +506,37 @@ impl<'a, P: StoreFamily<'a>> GFAStore<'a, P> {
...
+        let end_offset = compress_into_buffer(seq, &mut compressed);
+        let byte_span = self.seq_data.add_slice(&compressed);
+        let start = SeqSpan::to_logical(byte_span.start.index(), false);
+        let end = SeqSpan::to_logical(byte_span.end.index() - 1, end_offset) + 1;
+        self.segs.add(Segment {
+            name,
+            seq: SeqSpan {
+                start,
+                len: (end - start) as u16,
+            },
```
sampsyo: This would probably be a good place for using that `From<Range>` impl, if we do it.

sampsyo (commented):
Here are a few initial comments on the FlatGFA integration itself! I'll surely have more to say after we do a bit more disentangling of the other stuff from this PR, but this will get us started…

sampsyo:
I retitled this to better reflect is role as the successor to #214.

bench/__init__.py
sampsyo: Looks like diff noise?

pollen_py/pollen/depth/python_depth.py
sampsyo: Let's revert this whitespace change.

PR #242, opened 2025-12-10
Bencher Test Suite

johnpalsberg:
Adds a test suite for benchmarking performance and file size via Bencher.


tests/.gitignore
```diff
@@ -28,3 +28,5 @@ subset-paths/*.out
 depth/*.out
 depth/basic/*.out
 depth/subset-paths/*.out
+
+
```
sampsyo: Let's revert these changes, because they are kinda "diff noise."

filesize_benchmark.py
```diff
@@ -0,0 +1,32 @@
...
+
+def benchmark(test_file):
+  subprocess.run(["fgfa", "-I", test_file, "-o", "filesize_benchmark.txt"], check = True) 
+  size_bytes = os.path.getsize("filesize_benchmark.txt")
+  subprocess.run(["rm", "-rf", "filesize_benchmark.txt"], check = True) 
+  return size_bytes
+
+size_bytes_1 = float(benchmark("tests/chr6.C4.gfa")) / 1000.0
+size_bytes_2 = float(benchmark("tests/DRB1-3123.gfa")) / 1000.0
+size_bytes_3 = float(benchmark("tests/LPA.gfa")) / 1000.0
```
sampsyo: It seems like this will be a little easier to maintain if we make this more "data-driven," i.e., use a list for the filename and loop over it? Like this:

```
gfa_files = ["tests/chr6.C4.gfa", "tests/DRB1-3123.gfa", "tests/LPA.gfa"]
sizes = {name: float(benchmark(name)) / 1000.0 for name in gfa_files}
```

Then use that size dictionary in the Bencher results here.

filesize_benchmark.py
```diff
@@ -0,0 +1,32 @@
+import sys
+import os
+import json
+import subprocess
+
+subprocess.run(["cargo", "build", "--release"], check = True)
```
sampsyo: Maybe this is silly, but do we really want to do the cargo build here? It seems somewhat odd that the benchmarks script would do the rebuild itself rather than just using whatever had already been built. Furthermore, the fact that the responsibilities for "installing" `fgfa` are spread across two places (here and the GitHub Actions stuff, which amends `$PATH`) is a tad confusing.

I gently suggest that we remove this and instead add a step to the Actions script to do the build.

sampsyo (changes requested):
Great; thanks for getting this started! I have a few granular coding suggestions before we merge.

Also, can we please put these new scripts into a directory? We could reuse the `bench/` top-level directory that I made for my old infrastructure or pick a new, different name.

PR #250, opened 2026-03-27
cpp compatibility for header

nsmlzl:
@sampsyo we probably want cbindgen automatically add #ifdef __cplusplus / extern "C" guards. 

With this we should be able to just use it from both C and CPP.


sampsyo:
Good call; thanks!!

PR #252, opened 2026-04-23
All-Rust window depth example

sampsyo:
This is my attempt to extract @johnpalsberg's work on implementing the little odgi-documentation-shell-script example described in #206. I cherrypicked his commit 218305b5a199790631c0d18921d6e7fee87dca85, which appears to contain an MVP end-to-end example.

The main thing I then did was move this to the `examples/` directory, which is a special [feature in cargo](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples) just for showing off example uses of a library. I then took his primary test and turned it into a `main` function so you can invoke it from the command line. The result is it's possible to run this example tool like so:

```
cargo run --example window_depth <GFA> <BED> <CHROMOSOME> <WINDOWSIZE>
```

For example:

```
cargo run --example window_depth ../tests/k.gfa foo.bed x 5
```

That produces, for me:

```
x	0	5	2
x	5	10	1.5999999999999999
x	10	15	1.7999999999999998
x	15	20	2
x	20	25	2
x	25	30	2
x	30	35	1.7999999999999998
x	35	40	1.7999999999999998
x	40	45	2
x	45	50	2
```

Which I hope is correct?? At least the window indices look right.


sampsyo:
I also cherry-picked @johnpalsberg's more recent float-formatting fix from d7a31bb0c3eaaab7a9646807d9b39f47eecbf4ae. The output of that command above now looks like this:

```
x	0	5	2
x	5	10	1.6
x	10	15	1.8
x	15	20	2
x	20	25	2
x	25	30	2
x	30	35	1.8
x	35	40	1.8
x	40	45	2
x	45	50	2
```

